PR #3, opened 2026-02-17
CI: upload PostgreSQL logs on E2E failure

pinodeca:
Uploads PostgreSQL logs as an artifact when the E2E step fails (useful for diagnosing CI-only failures).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -81,7 +81,18 @@ jobs:
         run: cargo pgrx test pg17
 
       - name: Run E2E tests
+        id: e2e_tests
         run: ./scripts/test-e2e-local.sh
+
+      - name: Upload PostgreSQL logs on E2E failure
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -81,8 +81,19 @@ jobs:
         run: cargo pgrx test pg17
 
       - name: Run E2E tests
+        id: e2e_tests
         run: ./scripts/test-e2e-local.sh
 
+      - name: Upload PostgreSQL logs on E2E failure
```
kk-src, line 87: Do we need an ID as well added ?

kk-src (approved):

.github/workflows/ci.yml
pinodeca: I don't think so - an id is only needed to refer to the step elsewhere.

PR #31, opened 2026-03-03
[DO NOT MERGE] Add native Duroxide provider, stop using duroxide-pg-opt

pinodeca:
This is a POC/experiment to understand the feasibility of dropping the dependency on duroxide-pg-opt and using a native duroxide provider in pg_durable. The provider design would then be tailored to the PostgreSQL extension environment.


pinodeca:
Not going in this direction.

PR #36, opened 2026-03-04
ci: optimize caching and guard pgrx setup steps

pinodeca:
## Changes

- **Split single cache into two separate caches:**
  - **Cargo dependencies** (`~/.cargo/registry`, `~/.cargo/git`, `cargo-pgrx`, `target`) — keyed on `Cargo.lock` hash + PG version
  - **PostgreSQL build** (`~/.pgrx/config.toml`, `~/.pgrx/$PG_VERSION.*`) — keyed on PG version only (stable across dependency changes)

- **Exclude data directory from cache** — `~/.pgrx/data-*` (66 MB) was being cached with stale PG state, causing flaky tests. It only takes ~1s to recreate via `initdb`. An explicit removal step ensures no stale data leaks through.

- **Guard `cargo-pgrx` install** — skip if version 0.16.1 is already installed from cache (~30-60s saved)

- **Guard `pgrx init`** — skip PG download+compile if the build directory already exists (~2-5min saved)

## Why

Caching the PG data directory was the root cause of sporadic CI test failures. The `--clean` flag in E2E scripts was a workaround, but unit tests (`cargo pgrx test`) could still hit stale state. This eliminates the problem at the source.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,23 +98,46 @@ jobs:
               libxslt1-dev \
               libicu-dev
 
-      - name: Cache cargo registry
+      - name: Cache cargo dependencies
         uses: actions/cache@v4
         with:
           path: |
             ~/.cargo/registry
             ~/.cargo/git
-            ~/.pgrx
+            ~/.cargo/bin/cargo-pgrx
             target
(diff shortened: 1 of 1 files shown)


pinodeca:
This reduced the CI from 20+ minutes to 6.

PR #40, opened 2026-03-05
[DO NOT MERGE] Implement RLS and use SPI for PG backend interaction with duroxide

pinodeca:
Replace async sqlx/tokio calls with direct SPI in PostgreSQL backend
processes (user sessions). This eliminates TCP connection pools and
async runtime overhead for client, monitoring, and explain operations.
The background worker continues to use sqlx via duroxide-pg-opt since
the Provider trait requires Send + Sync + async.

This work builds on the POC in docs/design-native-provider.md, which
explored three phases: SPI for client ops, SPI for monitoring, and UDS
for BGW connections. This branch isolates the first two phases (SPI).

## Changes
- Rewrite `src/client.rs` to enqueue work items via SPI instead of async Client/Provider
- Rewrite `src/monitoring.rs` to query duroxide stored procedures via SPI
- Rewrite `src/explain.rs` `get_duroxide_instance_info()` to use SPI
- Add GRANTs on `df` and `duroxide` schema objects so non-superusers can call functions through SPI (which runs as the calling user)
- Enable Row-Level Security on `df.instances`, `df.nodes`, `duroxide.instances`, `duroxide.executions`, `duroxide.history`, and `duroxide.orchestrator_queue` so users only see/modify their own data
- Fix `21_signals` test: split signal+wait into separate transactions for SPI transaction visibility
- Add `docs/design-backend-spi.md` design document

### DIFF
--- a/docs/design-backend-spi.md
+++ b/docs/design-backend-spi.md
@@ -1,0 +1,356 @@
+# Backend SPI: Eliminate sqlx from PostgreSQL Backend Processes
+
+**Status:** In progress — SPI code done, permissions/RLS design needed
+**Branch:** `pinodeca/backend-spi`
+
+## Motivation
+
+pg_durable runs inside the PostgreSQL server as an extension. When a user calls
+`df.start()`, `df.status()`, or any other `df.*` function, that code executes in
+the user's PostgreSQL backend process.
+
+Previously, these backend functions created **TCP connections back to the same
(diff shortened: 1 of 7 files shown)


pinodeca:
#47 implements RLS

SPI for duroxide-pg-opt has been de-prioritized. Abandoning this PR.

PR #59, opened 2026-03-13
Validate df.break() is inside a loop at graph construction time

pinodeca:
## Problem

`df.break()` used outside a `df.loop()` body silently produced a corrupted result — the raw internal `{"__break__": true, "value": ...}` sentinel was written as the instance output, and the instance was marked `completed`. No error was raised.

## Fix

Reject such graphs immediately at `df.start()` time, before any nodes are inserted into the database.

### Changes

- **`src/types.rs`**: Add `Durofut::validate_structural(in_loop: bool)` which walks the graph, tracking whether execution is inside a loop body. A `BREAK` node encountered with `in_loop = false` returns a clear error. `LOOP` nodes recurse into their body with `in_loop = true` and into their while-condition with `in_loop = false`.

- **`src/dsl.rs`**: Call `validate_structural(false)` in `df.start()` after the existing `validate_recursive()` check, before any nodes are inserted.

- **`tests/e2e/sql/39_break_outside_loop.sql`**: E2E test covering four cases:
  1. Bare `df.break()` at top level → must error
  2. `df.break()` inside `~>` sequence without loop → must error
  3. `df.break()` in conditional branch outside loop → must error
  4. `df.break()` inside `df.loop()` body → must complete successfully

## Behavior table

| Pattern | Before | After |
|---|---|---|
| `df.start(df.break())` | Completed with sentinel result | Error at `df.start()` |
| `df.start('step1' ~> df.break())` | Completed with sentinel result | Error at `df.start()` |
| `df.start(df.loop('step' ~> df.break()))` | Completed correctly | Completed correctly |

### DIFF
--- a/src/dsl.rs
+++ b/src/dsl.rs
@@ -574,6 +574,10 @@ pub fn start(
     if let Err(e) = durofut.validate_recursive() {
         pgrx::error!("Invalid durable function graph: {}", e);
     }
+    // Validate structural constraints (e.g. df.break() must be inside df.loop())
+    if let Err(e) = durofut.validate_structural(false) {
+        pgrx::error!("Invalid durable function graph: {}", e);
+    }
     let instance_id = short_id();
 
     // Validate that the target database exists (if specified)
--- a/src/types.rs
+++ b/src/types.rs
@@ -538,6 +538,42 @@ impl Durofut {
         Ok(())
     }
 
+    /// Validate structural constraints that depend on nesting context.
+    ///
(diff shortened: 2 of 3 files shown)


pinodeca:
#229 catches break-outside-loop at runtime. This PR (#59) would fail-fast by catching it in df.start.

I'm going to close this without merging. We can re-open/re-visit fail-fast when either:
- users bring it up as helpful because they're accidentally using df.break outside loops
- OR we introduce a significant set of submission-time checks in df.start.

PR #80, opened 2026-03-24
Unify variable substitution syntax ($name for both named results and df.vars)

pinodeca:
Replace the dual-syntax variable substitution system (`{name}` for df.vars, `$name` for named results) with a single `$name` syntax for all variable types.

Function variables set via `df.setvar()` are now referenced as `$name` in queries, consistent with named results from `|=>`. The `{name}` syntax remains supported as a deprecated alias for backward compatibility.

## Changes

- **Proposal**: `docs/unify-vars.md` — analysis of why two syntaxes exist, benefits of unification, migration path
- **Implementation**: Extend `substitute_results()` to resolve `$name` against function variables as fallback
- **User Guide**: Updated variable substitution docs to use unified `$name` syntax
- **Backward compat**: `{name}` continues to work as a deprecated alias

### DIFF
--- a/docs/unify-vars.md
+++ b/docs/unify-vars.md
@@ -1,0 +1,166 @@
+# Proposal: Unified Variable Substitution Syntax
+
+**Status:** Draft  
+**Date:** 2026-03-24
+
+## Summary
+
+pg_durable currently uses two different syntaxes for variable substitution in SQL queries and HTTP templates:
+
+| Mechanism | Syntax | Source |
+|-----------|--------|--------|
+| Named results | `$var`, `$var.col`, `$var?`, `$var.*` | `\|=>` operator (step outputs) |
+| Function variables | `{var}` | `df.setvar()` / `df.vars` table |
+| System variables | `{sys_instance_id}`, `{sys_label}` | Runtime metadata |
+
+This proposal argues that named results and function variables should share the same `$` syntax, making `{var}` an alias for `$var` (or vice versa). **No changes to how named results are produced (`|=>`) or how function variables are set (`df.setvar()`) are proposed** — only the substitution syntax at query execution time.
+
+## Current Behavior
+
+### Named results (`$`)
+
+Created by the `|=>` operator during graph execution. The result of a SQL step is stored as a JSON object and substituted into downstream queries:
+
+```sql
+SELECT df.start(
+    'SELECT 42 AS amount' |=> 'total'
+    ~> 'SELECT $total * 2'
+);
+```
+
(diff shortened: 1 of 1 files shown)


pinodeca:
## Named Results (`$name`) — SQL-quoted

When a result from `|=>` is substituted into SQL, values are **type-aware and escaped**:

- **Strings** → single-quote escaped and wrapped: `'O''Brien'`
- **Numbers/Booleans** → inserted bare: `42`, `true`
- **NULL** → literal `NULL`
- **Row-set expansion** (`$name.*`) → generates `(VALUES (...)) AS name("col1", "col2")` with identifier-quoted column names

This happens in `format_value()` at types.rs. The quoting is controlled by the `for_sql` flag — SQL contexts get quoting, while raw contexts (HTTP URLs/headers) don't.

## Function Variables (`{varname}`) — **no quoting**

User variables from `df.setvar()` and system variables (`{sys_instance_id}`, `{sys_label}`) are substituted via **raw string replacement** — no escaping or quoting at all. See types.rs.

## Why they differ

Named results are **data** — they come from SQL query output and could contain arbitrary values including single quotes, so they must be escaped to produce valid SQL and prevent SQL injection between steps.

Function variables are **configuration** — they're intended to be SQL *fragments* (table names, expressions, URLs) set by the user who wrote the workflow. Quoting them would break legitimate uses like `{source_table}` → `my_schema.orders`. They're analogous to `psql` `\set` variables: the user controls both the template and the value, so escaping would be counterproductive.

pinodeca:
The quoting difference isn't a strong argument against unification. The substitution engine already knows where a value came from (results map vs vars map) and what context it's in (SQL vs HTTP). Even with unified `$name` syntax, you could still apply SQL quoting to results and raw insertion for variables — the dispatch logic wouldn't change.

The stronger arguments for keeping separate variable substitution syntax are:

1. **Ambiguity.** If a named result and a variable share the same name, unified `$name` needs a precedence rule. With `$name` vs `{name}`, there's no conflict — they coexist cleanly.

2. **Readability.** A reader of the workflow immediately knows `$requestbody` is a runtime query result and `{azure_function_key}` is pre-set configuration. That distinction disappears with unified syntax.

3. **Error messages.** Today, a typo in `$foo` means "named result not found" and a typo in `{foo}` means "variable not found." With unified syntax, the error has to guess which you meant, or give a less helpful "not found in results or variables."

In the [Azure Functions example](examples/azure-functions/sql/03_start_workflow.sql), the mix of `$requestbody`, `$doc.id`, and `{azure_function_base_url}` in the same `df.http()` call is actually a good illustration — you can immediately see which values come from prior query steps vs external config. That visual distinction has value.

PR #94, opened 2026-03-29
Initialize duroxide-pg-opt during Codespaces setup

pinodeca:
## Summary
- initialize the duroxide-pg-opt submodule during devcontainer onCreate setup
- document that Codespaces can authenticate to the private submodule via the configured repository permission
- degrade gracefully outside Codespaces when the private submodule is not accessible

## Validation
- bash -n .devcontainer/onCreateCommand.sh
- bash -n .devcontainer/postCreateCommand.sh


pinodeca:
Abandoning this PR. The Codespace repo permissions don't work during Prebuild/onCreateCommand. This PR will be superseded by #92 which will use a PAT instead so that duroxide-pg-opt submodule can be initialized and the pg_durable can be built - during Prebuild.

PR #199, opened 2026-06-01
ci: add pgspot check for extension SQL

tjgreen42:
Adds a pgspot CI gate that lints the SQL pg_durable ships (generated install SQL + non-frozen upgrade scripts) for `search_path`/privilege-escalation issues, and schema-qualifies the install DDL in `src/lib.rs` so it passes.


pinodeca (approved):
I have 2 concerns about [ci: add pgspot check for extension SQL by tjgreen42 · Pull Request #199 · microsoft/pg_durable](https://github.com/microsoft/pg_durable/pull/199)

1. The new CI job introduces significant pipeline code duplication.
2. Hardening lib.rs (via scoping) is awesome and great for new installations, but: should we have the 0.2.2->0.2.3 upgrade script modify existing functions and other objects (constraints) so that upgrades get the same benefits? The user-base will still be very small by the time this ships, but we'd probably do this if it were already a 1.x version, right?
    - __EDIT: Opus 4.7 tells me this won't add value because the hardening was NOT on bodies of the PL/pgSQL / SQL functions. The hardening in this PR only matters when the DDL runs, not DML.__

What do you think about the points above AND if you think either should be addressed, would you prefer to do that in the same PR or merge this one and then do followup PRs?

tjgreen42:
Thanks!

1. **CI duplication** — fixed in 96da3d0. Dropped the standalone job and folded the gate into the existing pg17 `Clippy & Tests` job (two steps after clippy), reusing its build instead of standing up a second pgrx environment. Trade-off: it no longer fails as fast as a `[format]`-only job, but it still runs before the test steps and removes ~75 lines of duplicated setup.

2. **Upgrade script** — agree with your edit, nothing to do. The qualification is source-text only; Postgres binds built-ins to OIDs at parse time, so existing installs already have byte-identical stored forms. An upgrade script would re-run CREATE OR REPLACE for zero catalog change. That's also why this needed no version bump.

pinodeca (approved):
LGTM

PR #201, opened 2026-06-02
Move duroxide schema

### LINKED ISSUE #175: Move PostgresProvider's schema out of "duroxide"
It may not be a good idea for pg_durable to reserve the "duroxide" schema. Users of duroxide + duroxide-pg may want to use "duroxide" schema for their application's durable function state.

pinodeca:
## Summary

Adds an initial design and planning document for moving pg_durable's internal duroxide provider schema away from the generic `duroxide` name for new installations.

The plan covers:
- backward compatibility for already-shipped Azure and open-source installs that use `duroxide`
- a proposed postmaster/superuser GUC for configuring the provider schema
- fresh-install, upgrade, drop/recreate, and GUC-change scenarios
- implementation phases, testing strategy, and open design questions

Refs #175


sql/pg_durable--0.2.2--0.2.3.sql
tjgreen42, line 1: We should have a `pg_durable--0.2.3.sql`, no?  (And I don't think we need `pg_durable--0.2.2.sql` or `pg_durable--0.1.1.sql` anymore.)

sql/pg_durable--0.2.2.sql
tjgreen42, line 1: Confused as to why we're introducing both this file (normally would come in PR stamping 0.2.2) and the upgrade to 0.2.3 (normally would come in PR stamping 0.2.3) in the same PR

sql/pg_durable--0.2.2.sql
```diff
@@ -0,0 +1,1047 @@
...
+-- Per-user scoping: each user has their own variable namespace.
+CREATE TABLE IF NOT EXISTS df.vars (
+    name TEXT NOT NULL,
+    value TEXT,
+    owner REGROLE NOT NULL DEFAULT pg_catalog.quote_ident(current_user)::pg_catalog.regrole,
+    PRIMARY KEY (owner, name)
+);
+
+-- Sentinel table: the background worker writes its epoch_id here after
+-- initialising.  If the extension is DROP-ed and re-CREATEd between
```
tjgreen42, line 76: Should be `initializing`

src/lib.rs
```diff
@@ -620,11 +620,26 @@ END $$;
 
 extension_sql!(
     r#"
--- The duroxide schema is created here so the extension owns it.
--- No IF NOT EXISTS: fails loudly if a duroxide schema already exists,
+-- The duroxide provider schema is created here so the extension owns it.
+-- No IF NOT EXISTS: fails loudly if a _duroxide schema already exists,
 -- preventing adoption of a potentially attacker-crafted schema.
```
tjgreen42, line 625: Need a brief argument here or elsewhere that we haven't introduced a security gap where an attacker could create a `duroxide` (legacy-named) schema and use that to shadow the `_duroxide` one.

docs/move-duroxide-schema.md
```diff
@@ -0,0 +1,230 @@
...
+- A `.so` upgrade that arrives **without** `ALTER EXTENSION pg_durable UPDATE` must continue to operate against the legacy `duroxide` schema (see "Selection algorithm" below).
+
+There is **no in-place migration path** from `duroxide` to `_duroxide` for an existing cluster that wants to adopt the new name while preserving engine state. The only supported "adopt the new name" path is `DROP EXTENSION pg_durable CASCADE` followed by `CREATE EXTENSION pg_durable`, which is a destructive reset of durable engine state. This is acknowledged as a deliberate non-goal of this work.
+
+## Design Overview
+
+Rather than a GUC, the selected provider schema is exposed by a small extension-owned SQL function:
+
+```sql
+CREATE FUNCTION df.duroxide_schema() RETURNS TEXT
```
tjgreen42, line 64: It's a little funky that we have this function that never gets called whose mere existence serves as a flag, but I guess it works.

tjgreen42 (approved):
Seems serviceable.  I did have some questions about the extension SQL files, maybe we just need a followup tidying pass.

tjgreen42:
One more comment: the upgrade tests are sort of simulated-upgrade tests right now as they don't build or run older .so binaries.  We should fix that.  (Can be done in a followup.)

sql/pg_durable--0.2.2--0.2.3.sql
pinodeca: pg_durable--0.2.3.sql is generated by pgrx during build, hence we don't need to materialize it.

pg_durable--0.2.2.sql is materialized only because it's the install script for the first Open Source version. Upgrade tests use it as a baseline.

In the same way, pg_durable--0.1.1.sql was the baseline for upgrade testing before we open sourced. I agree that it can be cleaned up now, but that should be done in a separate PR.

sql/pg_durable--0.2.2.sql
pinodeca: Please see my reply to the previous comment.

sql/pg_durable--0.2.2.sql
pinodeca: Ack. Will fix.

src/lib.rs
pinodeca: This code generates the install script for the current version. It creates schema _duroxide and fails if that already exists. If it succeeds, it proceeds to define function df.duroxide_schema which returns "_duroxide".

The ugprade script v0.2.2->v0.2.3, defines the same function to return "duroxide", in this way installations that started on v0.2.2 (or earlier) can continue using the "duroxide" schema.

The duroxide provider schema is never changed after creation, so shadowing is not possible.

docs/move-duroxide-schema.md
pinodeca: It's used twice in types.rs:
- in the function that resolves the schema name for the background worker's connections
- in the function that resolves the schema name for a regular backend's PostgresProvider connections (uses SPI, which our worker cannot)

PR #206, opened 2026-06-04
Default worker role to postgres instead of azuresu

pinodeca:
## Summary

The `pg_durable.worker_role` GUC defaulted to `azuresu`, the privileged admin role on Azure Database for PostgreSQL — Flexible Server. On any non-Azure PostgreSQL (vanilla, RDS, Cloud SQL, self-hosted) that role does not exist, so out of the box the background worker emitted "role does not exist" warnings and silently failed to process workflows until an operator overrode the GUC.

This changes the default to the vendor-neutral `postgres` role, which matches the Dockerfile and upgrade-test configuration and works on most setups.

## Changes

- `src/lib.rs` — `WORKER_ROLE` GUC default and the SQL install-script fallback now use `postgres`
- `src/types.rs` — `get_worker_role()` fallback and its doc comment updated
- Docs that cite the default value updated: `README.md`, `docs/rls.md`, `docs/security-review/workbook-data.md`, `docs/security-review/threat-model.dfd-lite.yaml`

All three runtime fallbacks (GUC default, SQL install fallback, `get_worker_role()`) are kept in sync.

## Notes

- Behavior change only, not a schema change — the worker role is read at runtime, not stored. The Dockerfile and `test-upgrade.sh` already pin `postgres`, so they are unaffected.
- `sql/pg_durable--0.1.1.sql` is intentionally left unchanged (historical version fixture for upgrade testing).
- No Azure-specific operator note was added, since only operators set this GUC.

## Validation

- `cargo build --features pg17` succeeds
- `cargo fmt -p pg_durable -- --check` passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ GRANT pg_durable_user TO app_backend, etl_service;
 > **Note:** `GRANT EXECUTE ON ALL FUNCTIONS` only applies to functions that exist when the grant runs. After upgrading pg_durable with `ALTER EXTENSION pg_durable UPDATE`, re-run `df.grant_usage('role')` (or re-issue the manual grants) so new functions are accessible.
 
 **Key points:**
-- The background worker role (`pg_durable.worker_role` GUC, default: `azuresu`) **must be a superuser** — it bypasses RLS to manage all users' instances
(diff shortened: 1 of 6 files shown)


tjgreen42 (approved):
LGTM

PR #225, opened 2026-06-11
Surface execution-history lookup failures in df.instance_executions (#168)

### LINKED ISSUE #168: df.instance_executions() can return no rows for completed instances
## Summary

`df.instance_executions(instance_id, 1)` can return an empty rowset for some instances that are otherwise reported as completed by `df.list_instances()` and aggregate metrics.

## Expected behavior

Every completed instance should have at least one execution row retrievable through `df.instance_executions(instance_id, 1)`, or the function should return an explicit error explaining why execution history is unavailable.

## Actual behavior

For a subset of completed instances, `df.instance_executions(instance_id, 1)` returns no rows. This makes it impossible to distinguish "no execution history exists" from "execution history lookup failed".

## Repro shape

1. Run enough workflows to produce multiple completed instances.
2. Enumerate completed instance IDs with `df.list_instances()`.
3. For each completed instance, call `df.instance_executions(instance_id, 1)`.
4. Observe that some completed instances return an empty rowset.

## Impact

Per-instance latency and execution-history aggregation can undercount completed work or silently drop instances from diagnostics.

crprashant:
## Summary

Fixes #168.

`df.instance_executions()` could return **no rows for a completed instance**, making "this instance genuinely has no execution history" indistinguishable from "the execution-history lookup silently failed."

## Root cause

The function swallowed *every* failure path into an empty rowset:

- building the async runtime
- connecting to the duroxide store (`new_backend_provider`)
- `list_executions`
- `get_execution_info`

Each was `Err(_) => return vec![]`. But duroxide-pg writes the `instances` row and the first `executions` row in the same transaction, so a completed instance **always** has at least one execution row. An empty result could therefore only mean a silently-swallowed lookup failure.

## Fix

- Propagate the runtime / provider / `list_executions` / `get_execution_info` failures and raise an explicit error (`pgrx::error!`) instead of returning an empty rowset. The error is raised *outside* `rt.block_on(...)` to avoid longjmp-through-tokio, matching the existing pattern in `dsl.rs` / `client.rs`.
- The legitimately-empty case — a non-existent or non-owned instance, gated by the existing RLS existence check — still returns an empty rowset, which is the correct response.
- Validate `limit_count >= 1` and cap it at `10000`, matching `df.list_instances`.

Issue #168 explicitly accepts "return an explicit error" as a valid resolution.

## Testing

- `cargo build --features pg17` — clean (only a pre-existing, feature-gated `extract_host` dead-code warning that is unrelated to this change and absent under CI's feature combo)
- `cargo clippy --no-default-features --features pg17,http-allow-test-domains -- -D warnings` — clean (0 warnings)
- `cargo pgrx test pg17` (`./scripts/test-unit.sh`) — **145 passed, 0 failed**, 16 ignored
- `./scripts/test-e2e-local.sh` — **27 passed, 0 failed**, including the new regression test `tests/e2e/sql/05_instance_executions.sql`

### DIFF
--- a/src/monitoring.rs
+++ b/src/monitoring.rs
@@ -192,6 +192,14 @@ pub fn instance_info(
 }
 
 /// Get the last N executions for an eternal durable function (loop).
+///
+/// Distinguishes "this instance genuinely has no execution history yet" (empty
+/// rowset) from "the execution-history lookup failed" (explicit error). The
+/// latter — failing to build the runtime, connect to the duroxide store, list
+/// executions, or fetch a specific execution's info — now raises an error
+/// instead of being silently swallowed into an empty rowset. A completed
+/// instance always has at least one execution row, so an empty result for one
+/// previously masked a real lookup failure. See issue #168.
 #[pg_extern(schema = "df")]
 pub fn instance_executions(
     instance_id: &str,
@@ -206,11 +214,18 @@ pub fn instance_executions(
         name!(output, Option<String>),
     ),
 > {
+    if limit_count < 1 {
+        pgrx::error!("limit_count must be at least 1");
+    }
+    let limit_count = limit_count.min(10000);
+
     let pg_conn_str = postgres_connection_string();
     let provider_schema = backend_duroxide_schema();
     let instance_id_owned = instance_id.to_string();
 
     // Ownership check: SPI goes through RLS, so non-owned instances are invisible.
+    // A non-existent or non-owned instance legitimately has no history to show,
+    // so an empty rowset (not an error) is the correct response here.
     let exists: bool = Spi::get_one_with_args(
         "SELECT EXISTS(SELECT 1 FROM df.instances WHERE id = $1)",
         &[instance_id.into()],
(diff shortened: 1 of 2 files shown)


pinodeca (approved):
LGTM; reviewed with Opus 4.8

### Minor / non-blocking

- **Test file naming:** tests/e2e/sql/05_instance_executions.sql duplicates the `05_` prefix of the existing 05_monitoring_and_explain.sql, which already tests the monitoring functions. Folding it in there (or using a fresh number) would be tidier. The repo tolerates duplicate numbers (three `22_` files) and the runner auto-discovers via `*.sql` glob into the "standard" phase, so it runs fine — purely cosmetic.
- The sibling `metrics()` function just below still uses the same `Err(_) => vec![]` swallow pattern. Out of scope for #168, but it carries the identical smell if you want a follow-up.
- The test's deliberate choice *not* to assert `status = 'completed'` (with the documented rationale about `df.instances` vs duroxide per-execution status diverging under BGW load) is well-reasoned and avoids a flaky race.

### Verdict

Approve. Solid, well-documented, minimal fix that matches issue #168's accepted resolution, with no security violations.

PR #229, opened 2026-06-12
Replace df.break() JSON sentinel with typed NodeError (#148)

### LINKED ISSUE #148: Refactor: propagate df.break() via a typed NodeError variant instead of a JSON sentinel
## Background

PR #140 fixes #132 by adding explicit `is_break_signal` checks inside `execute_join_node` and `execute_race_node`. That's the right point fix, but the underlying pattern is fragile and will keep producing the same class of bug.

Today, `df.break()` is encoded as an **in-band JSON sentinel** (`{"__break__": true, "value": ...}`) carried inside the same `Result<String, String>` that every node returns. Every compound node (`~>`, `if`, `join`, `race`, and any future construct like `try/catch`, `with_timeout`, `compensate`, …) must remember to:

1. Call `is_break_signal(&child_result)` after awaiting each child.
2. Early-return before doing its own post-processing (storing `result_name`, merging branch results, etc.).

The compiler offers no help. #132 happened because exactly this step was forgotten in two places. The next compound node added will likely reintroduce it.

There's also a minor collision risk: a user query returning JSON like `{"__break__": true}` would be misinterpreted as a break signal.

## Proposal

Encode break as a typed error variant and let `?` handle propagation:

```rust
enum NodeError {
    Break(String),      // unwinds until the nearest loop catches it
    Failure(String),    // real failure, propagates to orchestration result
}
type NodeResult = Result<String, NodeError>;
```

Then:

- **`execute_break_node`** returns `Err(NodeError::Break(value))` instead of the JSON sentinel.
- **`execute_then_node`, `execute_join_node`, `execute_race_node`, `execute_if_node`** become `child().await?` — propagation is automatic, no explicit `is_break_signal` checks.
- **`execute_loop_node`** is the only place that catches it:
  ```rust
  match execute_body(...).await {
      Ok(v) => { /* normal: check while-condition / continue_as_new */ }
      Err(NodeError::Break(v)) => return Ok(v),
      Err(e @ NodeError::Failure(_)) => return Err(e),
  }
  ```
- **`SubtreeEnvelope`** gains a typed field so a break inside `execute_subtree` is re-raised as `Err(Break(..))` in the parent rather than smuggled through the JSON result string.

## Benefits

1. **The bug class disappears.** Forgetting to propagate break becomes a type error — `?` handles it.
2. **No sentinel collision** — user payloads cannot impersonate control flow.
3. **`extract_break_value` / `BREAK_SENTINEL` JSON-munging goes away.** The value travels as a plain `String` field on the enum (no double serialization).
4. **JOIN multi-break semantics become explicit.** Today the first-in-iteration-order break silently wins; with a typed variant we can pick semantics deliberately.
5. **"`df.break()` outside a loop" becomes easy to detect.** It surfaces as a single uncaught `NodeError::Break` at the orchestration boundary, which we can convert to a clean error instead of returning a JSON sentinel as the function's final result (the current behavior).

## Costs

- Mechanical refactor across every node handler signature in `src/orchestrations/execute_function_graph.rs`.
- Small change to `SubtreeEnvelope` and the JOIN/RACE sub-orchestration boundary to carry the typed control-flow signal back to the parent.
- Need to update unit/E2E tests that may rely on the sentinel format (currently none surface it publicly, but worth verifying).

## Acceptance

- All existing E2E tests pass, including `tests/e2e/sql/22_break_in_join_race.sql` from #140.
- `is_break_signal`, `extract_break_value`, and `BREAK_SENTINEL` are removed.
- New test: `df.break()` used at the top level of `df.start()` (no enclosing loop) produces a clear, actionable error rather than completing with a sentinel as the result.
- New test: `df.break()` inside a nested compound (e.g., `df.if(..., df.break('x'), ...)` inside a JOIN inside a loop) correctly exits the outer loop.

## Related

- PR #140 — point fix for JOIN/RACE
- Issue #132 — original bug report

crprashant:
## Summary

Refactors `df.break()` propagation from an in-band JSON sentinel string
(`{"__break__": true, "value": ...}`) to a typed `NodeError::Break` enum, so Rust's
`?` operator auto-propagates break through every compound node (THEN / IF / JOIN / RACE
and the sub-orchestration boundary). The loop node is now the **sole** break catcher,
which turns "forgetting to propagate a break" from a recurring silent-bug class into a
compile error.

Fixes #148.

## Changes

- **Typed control flow**: new `NodeError { Break, Failure }` + `NodeResult`;
  `From<String>` / `From<&str>` map plain errors to `Failure` so activity/helper calls
  keep their `?` ergonomics.
- **Subtree boundary**: break now crosses the sub-orchestration boundary via a typed
  `control: Option<SubtreeControl>` field on `SubtreeEnvelope` instead of a sentinel
  smuggled in `result`. The `Option` makes an absent field (an envelope written by a
  `<= 0.2.2` binary) distinguishable from a new-binary `Some(Normal)`, so the legacy
  sentinel fallback runs *only* on genuinely old envelopes — a fresh JOIN/RACE branch
  result can never impersonate a break.
- **In-flight-upgrade safety**: when `control` is absent (`None`), `parse_subtree_envelope`
  re-raises a legacy `{"__break__"}` sentinel (written by `<= 0.2.2` binaries) as
  `NodeError::Break`, so a break persisted by an older binary is not silently swallowed
  across a binary upgrade.
- **Docs**: documented the Break-is-control-flow contract on the `execute()` entry point
  and in `docs/ARCHITECTURE.md`.
- **Tests**: 11 new unit tests for envelope parsing (legacy + typed, incl. a guard that a
  new-binary `Normal` result shaped like the legacy sentinel is *not* re-raised as a break);
  tightened the break-in-join-race E2E to assert the exact break value.

## Validation

- `cargo build --features pg17` — clean
- `cargo clippy --features pg17` — clean (only the pre-existing `extract_host` warning)
- `./scripts/test-unit.sh` — 156 passed, 0 failed
- `./scripts/test-e2e-local.sh` — 32 passed, 0 failed

## Upgrade notes

The only behavior change for already-running instances affects a **misuse** case: an
orchestration started under `<= 0.2.2` that hits a **top-level** `df.break()` (one used
outside any `df.loop()`) previously *completed* carrying the sentinel value; under the new
binary that uncaught break surfaces as a clear failure ("df.break() was called outside of a
loop"). JOIN/RACE-in-loop breaks persisted by an old binary still unwind correctly via the
legacy-sentinel fallback. No schema migration is required.

## Review

Reviewed via a multi-specialist pass (correctness, edge-cases, architecture,
maintainability, testing) plus synthesis. All three must-fix findings are addressed in
this PR; remaining should-fix / consider items were deferred per the synthesis
dispositions.

## Notes

A pre-existing, unrelated bug exists where a JOIN/RACE **inside a loop** running
>= 2 iterations can stall due to `continue_as_new` sub-orchestration id collisions. It is
independent of this change and is tracked separately in #230.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -521,6 +521,42 @@ pub async fn execute(
 
 ### Node Execution
 
+Internal node handlers return `NodeResult`, a `Result` whose error arm is a typed
+`NodeError` rather than a plain `String`. This lets `df.break()` propagate through the
+compound nodes (`THEN`, `IF`, `JOIN`, `RACE`) automatically via the `?` operator, instead
+of every handler having to recognise an in-band JSON break sentinel:
+
+```rust
+// src/orchestrations/execute_function_graph.rs
+pub enum NodeError {
+    /// df.break() fired. Carries the break value. Caught only by execute_loop_node.
+    Break(String),
+    /// A genuine failure. Surfaces as a failed instance.
+    Failure(String),
+}
+
+pub type NodeResult = Result<String, NodeError>;
+
+// Any `?` on an existing Result<_, String> auto-converts the error to Failure, so
+// activity calls and helpers need no per-call changes.
+impl From<String> for NodeError {
+    fn from(e: String) -> Self {
+        NodeError::Failure(e)
+    }
+}
+```
+
+`execute_loop_node` is the only handler that catches `NodeError::Break` (turning it into the
+loop's `Ok` result); `NodeError::Failure` keeps propagating. The orchestration boundary
+functions (`execute` / `execute_subtree`) still return `Result<String, String>` because they
+are registered with duroxide:
+
+- `execute`: an uncaught top-level `Break` becomes a clear `Err` ("df.break() was called
+  outside of a loop"), so the instance fails instead of completing with a sentinel value.
+- `execute_subtree` (used by JOIN/RACE branches): a `Break` is carried out-of-band in the
+  subtree envelope's `control` field and re-raised as `NodeError::Break` by
+  `parse_subtree_envelope` in the parent orchestration.
+
 The orchestration walks the graph recursively:
 
 ```rust
@@ -559,7 +595,7 @@ async fn execute_function_node_with_vars(
     node_id: &str,
     results: &mut HashMap<String, String>,
     exec_ctx: &ExecutionContext,
-) -> Result<String, String> {
+) -> NodeResult {
     let node = graph.nodes.get(node_id).ok_or("Node not found")?;
     
     ctx.trace_info(format!("Executing node {} (type: {})", node_id, node.node_type));
@@ -575,7 +611,7 @@ async fn execute_function_node_with_vars(
         "HTTP" => execute_http_node(ctx, node, results, exec_ctx).await?,
         "SIGNAL" => execute_signal_node(ctx, node).await?,
         "BREAK" => execute_break_node(ctx, node, node_id).await?,
-        _ => return Err(format!("Unknown node type: {}", node.node_type)),
+        other => return Err(NodeError::Failure(format!("Unknown node type: {other}"))),
     };
     
     // Store named results for $variable substitution
@@ -754,20 +790,21 @@ async fn execute_loop_node(
     node: &FunctionNode,
     results: &mut HashMap<String, String>,
     exec_ctx: &ExecutionContext,
-) -> Result<String, String> {
+) -> NodeResult {
     let body_id = node.left_node.as_ref().ok_or("LOOP missing body")?;
     
-    // Execute loop body
-    let body_result = execute_function_node_with_vars(
+    // Execute loop body. A df.break() anywhere in the body (including inside nested
+    // IF/JOIN/RACE) surfaces here as Err(NodeError::Break); the loop is the only node
+    // that catches it. NodeError::Failure keeps propagating out via `?`.
+    let body_result = match execute_function_node_with_vars(
         ctx, graph, body_id, results, exec_ctx
-    ).await?;
+    ).await {
+        Ok(v) => v,
+        Err(NodeError::Break(break_value)) => return Ok(break_value), // exit the loop
+        Err(e @ NodeError::Failure(_)) => return Err(e),
+    };
     
-    // Check for BREAK sentinel
-    if is_break_result(&body_result) {
-        return Ok(extract_break_value(&body_result));
-    }
-    
-    // Check while-condition if present
+    // Check while-condition if present (conditions are SQL and cannot break)
     if let Some(condition_node_id) = get_condition_node(node) {
         let condition_result = execute_function_node_with_vars(
             ctx, graph, &condition_node_id, results, exec_ctx
@@ -790,7 +827,7 @@ async fn execute_loop_node(
         .continue_as_new(serde_json::to_string(&new_input)?)
         .await
         .map(|_| body_result)
-        .map_err(|e| format!("continue_as_new failed: {:?}", e));
+        .map_err(|e| NodeError::Failure(format!("continue_as_new failed: {:?}", e)));
 }
 ```
 
--- a/src/orchestrations/execute_function_graph.rs
+++ b/src/orchestrations/execute_function_graph.rs
@@ -32,15 +32,74 @@ struct ExecutionContext {
     label: Option<String>,
 }
 
+/// Control-flow-aware error type returned by every node handler.
+///
+/// `Break` is **not** a failure: it unwinds through compound nodes (THEN, IF, JOIN,
+/// RACE, and the subtree boundary) via the `?` operator until the nearest enclosing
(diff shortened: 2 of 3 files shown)


pinodeca (changes requested):
Reviewed with Opus 4.8

### One real finding

**The legacy fallback re-introduces the exact sentinel-collision the refactor claims to eliminate, for the JOIN/RACE subtree path.**

In `parse_subtree_envelope`:

```rust
SubtreeControl::Normal => match parse_legacy_break_sentinel(&envelope.result) {
    Some(value) => Err(NodeError::Break(value)),
    None => Ok(envelope.result),
},
```

Because `control` uses `#[serde(default)]`, a **new-binary** normal branch (which writes `control = Normal`) is indistinguishable at parse time from an **old-binary** envelope (no field). So `parse_legacy_break_sentinel` runs on *every* normal subtree result, including fresh ones. A JOIN/RACE branch whose legitimate SQL result is shaped like `{"__break__": true, "value": ...}` would be falsely re-raised as a `Break` and exit the loop.

Issue #148 explicitly lists "user payloads cannot impersonate control flow" as benefit #2 — this is honored for the top-level/THEN/IF paths but silently reintroduced for subtree branch results. Low probability, but it's the same bug class, and it's a data-driven control-flow hijack rather than a crash.

**Concrete fix** — make field presence distinguishable instead of defaulting:

```rust
struct SubtreeEnvelope {
    #[serde(default)]
    control: Option<SubtreeControl>,  // None = pre-#148 binary
    result: String,
    results: HashMap<String, String>,
}
```

```rust
match envelope.control {
    Some(SubtreeControl::Break) => Err(NodeError::Break(envelope.result)),
    Some(SubtreeControl::Normal) => Ok(envelope.result), // new binary: NO legacy check
    None => match parse_legacy_break_sentinel(&envelope.result) {  // only old envelopes
        Some(value) => Err(NodeError::Break(value)),
        None => Ok(envelope.result),
    },
}
```

This keeps in-flight-upgrade safety (old envelopes still decode) while closing the collision for all envelopes written by the new binary. The existing tests would need their `control` arg adjusted, but the test helper already supports omitting the field for the legacy case.

### Minor notes

- Behavior change on resume: an orchestration started under ≤0.2.2 that later hits a *top-level* break will now *fail* instead of *complete* if it resumes under the new binary. This only affects the misuse case the PR intends to fix, but it's worth a one-line mention in the PR's upgrade notes.
- The two status-recording arms (`Ok` and `Err(Break)`) in `execute_function_node_with_vars` build a near-identical `completed` `update_node_status` payload. These could be collapsed by extracting the value first, then having a single `failed` vs `completed` branch. ~12 duplicated lines.

### Verdict

Solid, well-motivated refactor that achieves its primary goal with good attention to replay determinism and test coverage. I'd treat the **`SubtreeControl` collision via `#[serde(default)]`** as a should-fix before merge (simple `Option` change), since leaving it half-undercuts the stated security benefit. Everything else is polish.

crprashant:
@pinodeca thanks for the careful review — all three points are addressed in 5f2e9a6.

**Main finding (legacy fallback ran on every normal result).** You're right: with `#[serde(default)]` defaulting to `Normal`, a new-binary normal envelope was indistinguishable from a pre-#148 one, so `parse_legacy_break_sentinel` ran on every normal subtree result and re-opened the exact payload-impersonates-break collision this PR set out to close. Fixed by making the field `control: Option<SubtreeControl>`:

- `None` → envelope written by a `<= 0.2.2` binary (no `control` field) → run the legacy sentinel fallback.
- `Some(SubtreeControl::Normal)` → new binary → pass through untouched, **no** legacy check.
- `Some(SubtreeControl::Break)` → typed break.

I also dropped `Default`/`#[default]` from `SubtreeControl` since the default is no longer used. The wire format is unchanged — `Some(Normal)` still serializes to `"control":"Normal"` and an absent field still deserializes (now to `None`) — so replay across this change is byte-identical; the only behavior change is closing the collision. Added a regression test, `envelope_new_format_normal_with_sentinel_shaped_result_is_not_reraised`, which asserts a `Some(Normal)` envelope whose genuine result is shaped like `{"__break__": true, ...}` passes through as `Ok` rather than being re-raised.

**Minor — duplicated status-recording arms.** Collapsed the three near-identical arms in `execute_function_node_with_vars` into a single `(status, result)` match feeding one `update_node_status` schedule. Kept it to exactly one scheduled activity in every case so the recorded orchestration history stays identical for replay.

**Minor — resume behavior change.** Added an `## Upgrade notes` section to the PR description: an orchestration started under `<= 0.2.2` that hits a *top-level* break now surfaces a clear failure instead of completing with the sentinel value (JOIN/RACE-in-loop breaks from old binaries still unwind via the legacy fallback).

Validation after the change: `cargo fmt --check` clean, `clippy` clean (only the pre-existing `extract_host` warning), unit **156/0**, e2e **32/0**.

pinodeca (approved):
Thanks for addressing the issues raised - and for the fast turnaround!

LGTM - let's merge it!

PR #235, opened 2026-06-14
refactor: wire try_from_json() into production and clean up test memory

iemejia:
Follow-up to #220 addressing review feedback from @pinodeca:

1. **Wire `try_from_json()` into a production caller** — replaced direct `serde_json::from_str::<Durofut>` calls in `explain.rs` with `Durofut::try_from_json()`, so the fallible deserializer is no longer an unused parallel API.

2. **Test memory cleanup** — extracted a `build_wide_join(n)` helper that serializes the template node once and clones via `vec![value; n]` instead of calling `to_value()` per-iteration 10k times. Both node-count tests in `types.rs` and the `pg_test` in `lib.rs` now use this pattern for predictable memory.

Net: -18 lines, no behavioral changes.

### DIFF
--- a/src/explain.rs
+++ b/src/explain.rs
@@ -197,7 +197,7 @@ fn explain_expression(expr: &str) -> String {
     use crate::types::Durofut;
 
     // First try to parse as Durofut JSON
-    if let Ok(root) = serde_json::from_str::<Durofut>(expr) {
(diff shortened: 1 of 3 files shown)


pinodeca (approved):
Reviewed with Opus 4.8.

LGTM!

Issue #171, opened 2026-05-23
df.instance_nodes leaves race-loser nodes running or pending after race completion


pinodeca:
## Summary

After a `df.race(...)` workflow completes, `df.instance_nodes` can still show nodes from the losing branch as `running` or `pending`.

The underlying orchestration may have cancelled the losing branch correctly, but the operator-facing node view continues to look like work is still active.

## Expected behavior

Nodes in the losing branch of a completed race should be marked with a terminal status such as `cancelled` or `dropped`, consistent with the parent workflow having completed.

## Actual behavior

`df.status(instance_id)` reports the race workflow as completed, while `df.instance_nodes` still shows losing-branch nodes such as signal waits or downstream SQL nodes as `running` or `pending`.

## Repro shape

1. Start a workflow using `df.race(...)` with one branch that wins and another branch that waits or has downstream nodes.
2. Wait for the parent workflow to complete.
3. Query `df.instance_nodes` for the instance.
4. Observe losing-branch nodes still reported as active or pending.

## Impact

Diagnostics and dashboards that use `df.instance_nodes` to find stuck work can report ghost in-flight work for workflows that have already completed.

pinodeca:
This one is similar to #240 (addressed by WIP #249):
- After a failure, steps are skipped and should be marked accordingly (240)
- After a race is won, steps in the other branch are canceled (if they were running) or skipped

Do we need to be able to distinguish these - potentially 3 - cases?

pinodeca:
Addressed by #263 (node state-transition model).

`df.instance_nodes()` now returns a read-time **derived** column, `inferred_status`, in addition to the physical `status` column. The physical `status` of a race-loser node is unchanged — it can still read `pending` or `running`, because the worker doesn't go back and rewrite those rows. The fix is that diagnostics should read `inferred_status` (v0.2.4+) instead of `status`.

How it works: `infer_statuses()` walks the graph top-down from the root. When it reaches a `RACE` node that has reached a terminal status (the winner completed), every non-terminal node under the *losing* branch — signal waits, downstream SQL, etc. — is derived as `skipped`, with `inferred_status_from_ancestor_id` pointing back at the `RACE` node that decided it. So a completed race no longer shows ghost in-flight work when you query `inferred_status`.

Example:
```sql
SELECT node_id, node_type, status, inferred_status, inferred_status_from_ancestor_id
FROM df.instance_nodes('<instance_id>');
```

[...]

PR #249, opened 2026-06-18
WIP: start issue #240 implementation for skipped downstream node status

### LINKED ISSUE #240: No "skipped" indication for downstream steps after a step fails
## Problem

When a step fails, the workflow stops and remaining downstream steps do not execute. Their `df.nodes` rows (materialized up front by `df.start()`) remain in a `'pending'` state forever. There is no signal distinguishing:

- "this node was intentionally skipped because an earlier step failed", from
- "this node hasn't been reached yet because the run is still in progress".

The `df.nodes` status check constraint currently allows only:

```
'pending', 'running', 'completed', 'failed'
```

(see [src/lib.rs](src/lib.rs#L251)). There is no `'skipped'` terminal state, so a consumer inspecting a failed run's nodes cannot tell skipped-due-to-upstream-failure from never-reached.

## Expected

Either:

- (a) `df.nodes` rows for un-executed steps carry a distinct `'skipped'` status when the workflow terminates, or
- (b) a separate `df.pipeline_run_steps` projection materializes the full ordered step list per run with each step's terminal status (including `'skipped'`).

## Notes

- Severity: Low-Medium.
- Tracked internally as ADO 5317226 ([A3]).

Pyrotechnic679:
## Summary

This PR is an initial WIP for issue #240.

It starts implementation work to make downstream, unexecuted nodes observable after workflow failure (instead of remaining ambiguous as `pending`).

## Current state

- Initial schema/runtime/test/doc changes are included.
- Work is **not finished** yet.
- Full validation/testing has **not** been completed yet.

## Intent

Opening early to signal active work and allow maintainers to review direction before finalizing implementation details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,14 @@ Pre-1.0 note: while `pg_durable` is in major version `0`, minor releases may inc
 
 ## [0.2.4] - Unreleased
 
+### Added
+
+- **Skipped node terminal status:** downstream nodes that are not executed after an upstream node failure are now marked `skipped` instead of remaining indefinitely `pending` (#240).
+
 ### Changed
 
 - **`df.grant_usage()` / `df.revoke_usage()`:** dropped the explicit per-function `EXECUTE` allowlist. Schema `USAGE` on `df` is the real access gate for ordinary `df.*` functions, so the helpers now grant/revoke schema `USAGE`, the table privileges, and `EXECUTE` only on the sensitive functions (`df.http`, `df.grant_usage`, `df.revoke_usage`). Function signatures are unchanged and existing privileges are unaffected (#242).
+- **Failure observability:** on node-level workflow failure, pg_durable now performs a terminal reconciliation pass to mark remaining `pending` nodes as `skipped` when supported by the installed schema.
 
 ### Removed
 
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -1300,6 +1300,7 @@ SQL |=> 'a': SELECT 1
 | `✗ Failed` | Node encountered an error |
 | `⏳ Running` | Node currently executing |
 | `○ Pending` | Node waiting to execute |
+| `⊘ Skipped` | Node was not executed because an upstream node failed |
 
 ### Visualizing Complex Structures
 
--- a/docs/issue-240-skipped-status-plan.md
+++ b/docs/issue-240-skipped-status-plan.md
@@ -1,0 +1,168 @@
+# Issue 240 Design and Implementation Plan
+
+## Summary
+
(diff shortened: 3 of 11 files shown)


Pyrotechnic679:

pinodeca:
Reviewed and pushed a follow-up commit (`8d104c9`) addressing the main blocker plus some cleanup, and ran the local validation matrix.

### What changed
- **Upgrade-script DDL (blocking fix):** the WIP added `skipped` to `nodes_status_chk` in the fresh-install DDL (`src/lib.rs`) but not to the `0.2.3 → 0.2.4` upgrade script. Without it, **upgraded** installs keep the old constraint, so the runtime's schema probe finds `skipped` unsupported and silently no-ops forever — the feature would only work on fresh installs. Added a `DROP`/`ADD CONSTRAINT nodes_status_chk ... 'skipped' ... NOT VALID` block to `sql/pg_durable--0.2.3--0.2.4.sql` that matches the install DDL exactly (pure widening, no data migration, Scenario B2 safe).
- **E2E test cleanup (`tests/e2e/sql/49_...`):** switched to the canonical `df.await_instance` (every other e2e test uses it; `df.wait_for_completion` is the deprecated alias), fixed branch-filter indentation, and added the missing trailing newline.

### Validation run locally
- `cargo clippy --features pg17` — clean
- `./scripts/test-unit.sh` — 183 passed, 0 failed
- `./scripts/test-e2e-local.sh 49_failed_downstream_nodes_skipped` — PASS (both the THEN-chain and IF-condition scenarios)

### Still open / discussion
- **Schema-parity nuance:** `test-upgrade.sh` Scenario A only compares constraint *name/type*, not the CHECK expression, so it would not have caught the missing DDL on its own — the gap was purely functional. Worth keeping in mind.
- **Merge conflict ahead:** `main` has since added a pagination section to this same `0.2.3 → 0.2.4` script. A rebase will conflict in that file; resolution is to keep both sections.
- **Design question (relates to #171):** "not executed" really covers several distinct cases — upstream-failure skip (#240), conditional-branch-not-taken, and race-loser (#171, where the node may have been `running`, so it's arguably `cancelled` not `skipped`). We're going to flesh out an explicit node/instance state-transition model (leaning on Airflow's `skipped` vs `upstream_failed` distinction, with BPMN "withdrawn" semantics for race-losers) plus a `status_reason` so #240 and #171 are solved coherently together. More to come.

The orchestration change also adds a scheduled activity on the failure path; note this alters recorded history for any orchestration that fails across a binary upgrade (narrow window) — flagging for awareness.

pinodeca:
Superseded by #263 (node state-transition model), which addresses #240 with a read-time derived `inferred_status` column on `df.instance_nodes()`. Closing this WIP in favor of that approach.

See #240 for details on how #263 resolves the underlying issue (downstream nodes after a failure derive to `skipped` only when a terminal ancestor decided against them, while genuinely not-yet-reached nodes stay `pending` — no new physical status or constraint change needed).

PR #251, opened 2026-06-18
fix: isolate epoch/extension polling onto dedicated pool

iemejia:

Add a separate 1-connection `poll_pool` used exclusively for extension-existence checks and epoch sentinel heartbeats. This prevents activity work (graph loading, status updates) on the management pool from starving the health-check loop and causing spurious runtime shutdowns under load.

**Files:** `src/worker.rs`
**Tests:** Runtime wiring change — no unit-testable logic added.

### DIFF
--- a/src/worker.rs
+++ b/src/worker.rs
@@ -125,8 +125,7 @@ async fn run_duroxide_runtime() {
     );
 
     // Management pool: consolidates former polling and activity pools into one.
-    // Used for extension-existence polling, epoch sentinels, worker-ready writes,
-    // graph loading, and status updates. Sized by the max_management_connections GUC.
+    // Used for graph loading and status updates. Sized by the max_management_connections GUC.
     // Retry in a loop so the worker survives the target database not yet existing
     // (e.g. pg_regress creates `contrib_regression` after PostgreSQL starts).
     let mgmt_pool = loop {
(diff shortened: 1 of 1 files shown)


pinodeca (approved):
Reviewed with Opus 4.8

### Issues / suggestions

1. **Stale comment on `mgmt_pool` (should fix).** The revised comment now reads "Used for graph loading and status updates," but `mgmt_pool` is still used for more than that — `resolve_duroxide_schema_pool`, `initialize_duroxide_runtime`, and `write_worker_ready`. The comment is now inaccurate. Consider: "Used for schema resolution, runtime initialization, worker-ready writes, graph loading, and status updates."

2. **Connection-budget accounting (worth noting).** The worker logs a budget of `management + duroxide + user` (defaults 6 + 10 + 10). This PR adds a permanent +1 connection that isn't reflected in that log line or any `max_connections` sizing guidance. Peak usage is now `mgmt + duroxide + user + 1`. Minor, but since the original comment notes the pools were *consolidated* to simplify the budget, re-introducing a separate pool reopens that accounting — worth a line in the budget log or docs.

3. **Health check is still fragile to transient errors (likely out of scope, but flagging).** Even with the dedicated pool, `check_epoch_sentinel` treats *any* `Err` as "dropped." A transient connection blip or a brief DB restart will still cause a spurious shutdown. The dedicated pool removes the *acquire-timeout* cause but not the general class. A more robust fix would distinguish "row missing / undefined_table" (real drop) from `PoolTimedOut`/connection errors (transient → keep running). Reasonable to defer to a follow-up since this PR is a focused split from #221.

4. **Duplicated pool-creation retry loop (nit).** The `poll_pool` creation block is a near-copy of the `mgmt_pool` block. A small `create_pool_with_retry(conn_str, max_conns, label)` helper would remove the duplication, but it's acceptable as-is.

### Verdict
Approve with minor changes — the core fix is sound. I'd ask for the comment correction (#1) before merge, and at minimum a note acknowledging the extra connection (#2). Items #3 and #4 are fine as follow-ups.

PR #253, opened 2026-06-18
fix: block CGNAT range 100.64.0.0/10 in SSRF protection

iemejia:

Add RFC 6598 Carrier-Grade NAT addresses to the IPv4 blocklist. These are used by cloud providers for internal routing and could expose metadata endpoints via DNS rebinding attacks.

Also narrow the `cfg` gate on `extract_host` to match only the feature flags that actually use it (`http-allow-azure-domains`, `http-allow-test-domains`) instead of the overly broad `not(http-allow-all)`.

**Files:** `src/ssrf.rs`, `src/lib.rs`
**Tests:** 1 unit test + 1 pg_test

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2714,6 +2714,27 @@ mod tests {
             "Instance not found"
         ));
     }
+
+    // --- H6: CGNAT SSRF blocklist ---
+
+    #[pg_test]
+    fn test_ssrf_blocks_cgnat_range() {
+        use std::net::{IpAddr, Ipv4Addr};
+        // 100.64.0.0/10 must be blocked
+        assert!(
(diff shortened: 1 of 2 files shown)


pinodeca (approved):
Reviewed with Opus 4.8:

## Findings
1. **Medium — ungated `pg_test`:** `test_ssrf_blocks_cgnat_range` in lib.rs lacks `#[cfg(not(feature = "http-allow-all"))]`, unlike its sibling unit test in ssrf.rs. Under `http-allow-all`, SSRF returns `None` and this test would fail.
2. **Low — docs:** http-security.md §4.1 "Blocked IPv4 ranges" table wasn't updated to list `100.64.0.0/10` (RFC 6598).

pinodeca (approved):
LGTM!

PR #256, opened 2026-06-18
fix: wrap std::env::set_var in unsafe block for Rust 2024 edition

iemejia:

`std::env::set_var` becomes unsafe in the Rust 2024 edition because it can cause undefined behavior when other threads read the environment concurrently.

Add `SAFETY` documentation explaining why `set_var` is acceptable here: PostgreSQL backends are forked single-process, and the BGW tokio runtime uses `new_current_thread()` with no additional OS threads spawned.

**Files:** `src/worker.rs`, `src/client.rs`
**Tests:** Purely mechanical annotation — zero behavioral change.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -104,7 +104,13 @@ where
         let client = rt.block_on(async {
             // Limit backend provider to 1 connection — backends need minimal
             // duroxide access (start/cancel/signal only).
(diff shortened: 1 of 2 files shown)


pinodeca (approved):
LGTM!

pinodeca (approved):
LGTM!

PR #257, opened 2026-06-18
Fix pg_durable documentation examples

tjgreen42:
## Problem

Testing the README and user-facing documentation examples found several result-substitution issues:

- The README quick example used a captured multi-row result as a scalar in an array context. A bare `$var` capture substitutes only the first column of the first row, so the generated SQL was invalid.
- `USER_GUIDE.md` had additional examples with the same scalar-vs-row-set issue, unqualified playground tables, duplicate unnamed substituted columns, and signal workflows that captured `order`/`doc` but referenced undefined `$order_id`/`$doc_id`.
- `docs/spec-signals.md` duplicated the signal capture-reference mistakes.
- The Azure HTTP domains smoke check comment claimed broader SQL validation than the script performs.

## Fix

- Use `$batch.*` row-set expansion where examples pass captured row sets between steps.
- Keep runnable User Guide examples aligned with the `playground` schema from the appendix.
- Alias substituted dot-notation columns in the dot-notation example.
- Reference captured result columns with `$order.id` and `$doc.id` in signal examples.
- Narrow the smoke-check comment to match its actual offline validation.

## Verification

Tested against pgrx-managed PostgreSQL 17 and 18 instances built from this PR branch:

- README quick example completed and updated all expected rows.
- Corrected finite User Guide examples completed, including dot notation, `df.if_rows()`, configurable ETL, row-set expansion, and signal timeout flows.
- `docs/SCENARIOS.md` copy-paste scenarios completed; the scheduled HTTP scenario was started, observed through its first insert, then cancelled.
- Example smoke checks passed for `azure-functions`, `azure-http-domains`, and `invoice-approval`.
- Local non-cloud portions of the Azure example SQL scripts ran successfully.
- Patched signal snippets in `docs/spec-signals.md` completed with expected state changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ The model is intentionally SQL-shaped. If a step needs arbitrary code, a non-HTT
 -- A durable function that processes data in steps
 SELECT df.start(
     'SELECT id FROM documents WHERE processed = false LIMIT 100' |=> 'batch'
-    ~> 'UPDATE documents SET processed = true WHERE id = ANY($batch)'
+    ~> 'UPDATE documents SET processed = true WHERE id IN (SELECT id FROM $batch.*)'
 );
 ```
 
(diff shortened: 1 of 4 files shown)


pinodeca (approved):
Thanks for catching this!

pinodeca (approved):
LGTM! Thanks for QA-testing our docs!

PR #258, opened 2026-06-19
Prevent df/duroxide drift with transactional enqueue

### LINKED ISSUE #213: Rolled-back df.start() leaves failed duroxide.executions rows that inflate df.metrics()
## Summary

`df.metrics()` appears to count failed rows from `duroxide.executions`, including orphan executions created when `df.start()` is rolled back, while `df.instances` and `df.list_instances('failed')` only show persisted workflow instances.

That makes the failed instance count disagree across public APIs after rollback scenarios.

## Observed

Tested against current `main` at `11ac64e3adb64c14386be5c737b3a3806d873fc4`.

After rollback-oriented tests, the counts diverged:

```text
source               total  completed  failed  running
df.metrics()         399    392        7       0
df.instances         396    392        4       0
duroxide.executions  399    392        7       0
```

The extra failed rows were in `duroxide.executions` with no matching row in `df.instances`:

```sql
SELECT
  e.instance_id,
  e.execution_id,
  e.status,
  left(e.output, 180) AS output_prefix,
  i.id AS df_instance_id
FROM duroxide.executions e
LEFT JOIN df.instances i ON i.id = e.instance_id
WHERE e.status = 'Failed'
  AND i.id IS NULL
ORDER BY e.instance_id;
```

Example output prefix:

```text
Instance <id> not found after 5s (transaction may have been rolled back)
```

So `df.metrics()` reports these as failed instances even though `df.instances` and `df.list_instances('failed')` do not expose them as failed workflow instances.

## Repro Shape

One way to trigger this is to start a workflow inside a transaction that later rolls back, wait for the worker to observe the missing instance, then compare the metrics API with `df.instances`.

```sql
BEGIN;
SELECT df.start('SELECT 1', 'rollback-metrics-probe');
ROLLBACK;

-- wait long enough for the worker to record the missing instance failure

SELECT * FROM df.metrics();

SELECT status, count(*)
FROM df.instances
GROUP BY status;

SELECT e.instance_id, e.status, e.output, i.id AS df_instance_id
FROM duroxide.executions e
LEFT JOIN df.instances i ON i.id = e.instance_id
WHERE e.status = 'Failed'
  AND i.id IS NULL;
```

## Expected

Either:

- `df.metrics()` should count the same persisted workflow instances that `df.instances` / `df.list_instances()` expose, or
- the docs should clearly state that `df.metrics().failed_instances` includes lower-level failed `duroxide.executions`, including orphan executions created by rolled-back starts.

For dashboards and alerting, the current behavior makes rollback probes look like durable workflow failures.

## Notes

From a quick source read, `df.metrics()` appears to come from the generated `get_system_metrics()` path and counts failed rows in `duroxide.executions`. That explains why it can diverge from `df.instances` after rollback.

tjgreen42:
## Summary

Prevention half stacked on #279.

- Moves df.start, df.cancel, and df.signal to in-transaction enqueue via private SECURITY DEFINER wrappers when the duroxide-pg SQL surface is present.
- Keeps the provider-agnostic out-of-band fallback for schemas/providers without that SQL surface.
- Hardens wrapper authorization so callers cannot forge starts, cancels, or signals against other users instances.
- Requires start-wrapper enqueue to target a pending row inserted in the current transaction/subtransaction, preserving SET ROLE and SECURITY DEFINER df.start semantics while blocking cross-owner committed pending starts.
- Adds rollback/authorization E2E coverage, including cross-owner pending-start denial.
- Updates signal tests for the new early-signal behavior.

## Stack

1. #279 - repair/backstop
2. #258 - prevention via transactional enqueue

## Validation

- cargo fmt -p pg_durable -- --check
- cargo clippy --no-default-features --features pg17,http-allow-test-domains -- -D warnings
- ./scripts/test-e2e-local.sh 13_user_isolation --clean --default-build-phases --pg-version 17 --verbose
- ./scripts/test-e2e-local.sh 24_atomic_rollback --clean --default-build-phases --pg-version 17 --verbose
- ./scripts/test-e2e-local.sh 25_enqueue_wrapper_authz --clean --default-build-phases --pg-version 17 --verbose
- targeted signal/cross-connection/reconcile E2E
- full PG17 E2E, upgrade tests, and pg_regress before the split; CI reruns on this branch


sql/pg_durable--0.2.3--0.2.4.sql
```diff
@@ -166,3 +196,286 @@ CREATE FUNCTION df."await_instance"(
...
+
+    -- Authorization. This runs as the (privileged) definer, so it must not
+    -- trust the caller to only target their own instance. Permit the enqueue
+    -- only for a brand-new, not-yet-started instance: a 'pending' df.instances
+    -- row with no orchestrator-queue entry and no duroxide instance. Under the
+    -- atomic-start path a committed df.instances row always has its queue row in
+    -- the same transaction, so this state is reachable only for the instance the
+    -- caller (df.start) just inserted in the current transaction — never another
+    -- user's committed instance. This is what stops a caller from forging work
+    -- against a foreign instance.
```
pinodeca: Comment nit, and to check my understanding...

Given the permissions, a user can bypass df.start and populate df.instances and df.nodes tables, then call this function. The checks you are doing therefore limit the user to enqueueing any pending instance that they own, not necessarily one that's in the same transaction.

tjgreen42: Yes, that understanding is correct. The old comment overstated the guarantee: the wrapper permits a same-owner pending/not-yet-started instance, not necessarily one inserted by df.start in the same transaction. I updated the comment in both fresh-install SQL and the upgrade script to say that explicitly, and to point at the actual safety properties: fixed root orchestration, matching input instance id, and no foreign already-started target.

tjgreen42:
Closing as superseded. The transactional-enqueue approach (SPI into duroxide's internal schema) is not the direction we're taking — Affan's point that the duroxide schema is an internal detail, not the API, stands. df.start() now fail-fasts on a failed engine hand-off (#282), and orphaned engine state is reaped through duroxide's Rust API (#283) rather than by coupling to its internal schema.

PR #259, opened 2026-06-19
Create schemas, tables and indexes without "if not exists"

pinodeca:
## Summary

Eliminates silent schema adoption and `IF NOT EXISTS` from the install path, following the pg_cron anti-squatting model.

- Control file now uses `schema = pg_catalog`; the `df` and `_duroxide` schemas are created in a pgrx `bootstrap` block with plain `CREATE SCHEMA` (no `IF NOT EXISTS`). A pre-existing/squatted `df` schema now makes `CREATE EXTENSION` fail loudly (`duplicate_schema`) instead of being silently adopted.
- `df.duroxide_schema()` moved into the bootstrap block; the separate `create_duroxide_schema` block is removed.
- Removed `IF NOT EXISTS` from install-script tables (`df.nodes`, `df.instances`, `df.vars`, `df._worker_epoch`) and indexes so install fails loud on conflicts.
- `#[pg_schema] mod df {}` retained (required so `#[pg_extern(schema = "df")]` functions resolve). Its redundant `CREATE SCHEMA IF NOT EXISTS df` is now a no-op after bootstrap; the pgspot PS010 allowlist entry stays with an updated explanatory comment.

## Why not drop the PS010 exception?
pgrx's `#[pg_schema]` unconditionally emits `CREATE SCHEMA IF NOT EXISTS df` into the generated install SQL, and that string isn't configurable from source. Removing it would require post-processing the generated SQL, so the (now no-op) line and its single-line PS010 allowlist entry remain.

## Testing
- Unit: 183 passed
- E2E suite: 33 passed, 0 failed (incl. `12_extension_lifecycle` anti-squat Test 2)
- Upgrade tests: 36 passed (B1 backward-compat vs v0.2.2/v0.2.3 schemas, B2 data-compat, Scenario A schema comparison)
- `cargo fmt --check`, `cargo clippy`: clean
- pgspot gate: PASS (install SQL 0 errors / 0 warnings)

### DIFF
--- a/pg_durable.control
+++ b/pg_durable.control
@@ -4,15 +4,8 @@
 comment = 'pg_durable: SQL-native durable orchestrations for PostgreSQL'
 default_version = '@CARGO_VERSION@'
 module_pathname = 'pg_durable'
+schema = pg_catalog
+# pgrx's control-file parser requires `relocatable` and `superuser` to be
+# present explicitly (they are otherwise PostgreSQL defaults).
 relocatable = false
 superuser = true
-trusted = false
-# Note: 'schema' is intentionally omitted. This extension manages two schemas
-# (df and the duroxide provider schema), and PostgreSQL's control file only
-# supports a single schema directive. The df schema is created by pgrx
-# (#[pg_schema]); the duroxide provider schema is created by the
-# create_duroxide_schema extension_sql block in src/lib.rs (named '_duroxide'
-# on fresh installs, or the legacy 'duroxide' on installs upgraded from
-# <= 0.2.2). relocatable = false prevents schema relocation attacks.
-
-
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -142,7 +142,35 @@ pub extern "C-unwind" fn _PG_init() {
 // Schema Declaration
 // ============================================================================
 
-/// The 'df' schema contains all pg_durable functions (df = durable functions)
+// Create both extension-owned schemas as the very first statements of the
+// install script. `bootstrap` guarantees this runs before every other extension
+// object, including the redundant `CREATE SCHEMA IF NOT EXISTS df` that pgrx
+// emits for the `#[pg_schema] mod df` entity below.
+extension_sql!(
+    r#"
+CREATE SCHEMA df;
+CREATE SCHEMA _duroxide;
+
+-- Returns the name of the duroxide provider schema selected for this install.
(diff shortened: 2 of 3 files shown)


src/lib.rs
```diff
@@ -142,7 +142,35 @@ pub extern "C-unwind" fn _PG_init() {
...
 
-/// The 'df' schema contains all pg_durable functions (df = durable functions)
+// Create both extension-owned schemas as the very first statements of the
+// install script. `bootstrap` guarantees this runs before every other extension
+// object, including the redundant `CREATE SCHEMA IF NOT EXISTS df` that pgrx
+// emits for the `#[pg_schema] mod df` entity below.
+extension_sql!(
+    r#"
+CREATE SCHEMA df;
+CREATE SCHEMA _duroxide;
```
tjgreen42, line 152: I'm a little confused, this runs even for an upgrade and then we'll have both duroxide and _duroxide schemas?  (I guess this is preexisting behavior, but still seems weird...)  But it's kind of hard to tell what the final .sql scripts look like.

Followup: WDYT about checking in the generated install/upgrade scripts to make confirming this kind of thing easier?

pg_durable.control
```diff
@@ -4,15 +4,8 @@
 comment = 'pg_durable: SQL-native durable orchestrations for PostgreSQL'
 default_version = '@CARGO_VERSION@'
 module_pathname = 'pg_durable'
+schema = pg_catalog
+# pgrx's control-file parser requires `relocatable` and `superuser` to be
+# present explicitly (they are otherwise PostgreSQL defaults).
 relocatable = false
-superuser = true
-trusted = false
```
tjgreen42, line 9: Did you mean to get rid of the `trusted` line?

tjgreen42 (approved):

pg_durable.control
pinodeca: Yes, I removed lines with default values except those explicitly required by pgrx.

src/lib.rs
pinodeca: lib.rs only controls the install/create script, not the upgrade scripts, which are managed by hand in the sql folder (along with old install script kept for testing purposes only).

I agree that pgrx makes it hard to tell what the final .sql script, but we do check it in the sense that we have an upgrade test validate that install version X generates the same objects as installing version X-1 and then upgrading to X.

PR #265, opened 2026-06-24
fix: prune old terminal instances

### LINKED ISSUE #261: Periodically prune the instance and node tables
Periodically prune old terminal instances (and their corresponding rows in df.nodes). Hard coded 30 day expiration AND 10K (100K ?) instances should be fine for a first implementation (no GUC).

nanookclaw:
## Summary

Closes #261.

This adds an hourly best-effort maintenance pass to the background worker that prunes old terminal workflow history from `df.instances` and the matching `df.nodes` rows. The retention policy is intentionally fixed for now: keep terminal rows for at least 30 days, and always keep the newest 10,000 terminal instances even if they are older.

The pruning query runs in one transaction with constraints deferred, ranks only terminal instances (`completed`, `failed`, `cancelled`) by `COALESCE(completed_at, updated_at, created_at)`, deletes matching nodes first, then deletes the selected instance rows. Failures are logged and do not stop the worker runtime.

A focused pgrx test covers age cutoff, keep-count behavior, nonterminal retention, cancelled/failed/completed statuses, and node cleanup.

## Validation

- `cargo fmt --check`
- `git diff --check`
- `cargo check --lib` could not complete in this environment because pgrx build setup is unavailable: `$PGRX_HOME does not exist`.

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -1520,6 +1520,39 @@ SELECT started_at, last_seen_at,
 
 The background worker updates `last_seen_at` every ~5 seconds as part of its normal operation.
 
+### Data Retention (Automatic Pruning)
+
+To keep `df.instances` and `df.nodes` from growing without bound, the background
+worker runs a **best-effort pruning pass roughly once an hour** that deletes old
+**terminal** instances (status `completed`, `failed`, or `cancelled`) and their
+associated `df.nodes` rows. Running and pending instances are **never** pruned,
+regardless of age.
+
+The policy is currently fixed (no configuration GUC):
+
+- **Hard cap — at most 10,000 terminal instances are retained, regardless of
+  age.** The newest 10,000 terminal instances are kept; any beyond that are
+  pruned even if they are only minutes old.
+- **Retention window — 30 days.** Terminal instances older than 30 days are
+  pruned even if the table holds fewer than 10,000 of them.
+
+Equivalently, a terminal instance is retained only while it is **both** among the
+newest 10,000 terminal instances **and** less than 30 days old; otherwise it is
+eligible for pruning.
+
+Notes:
+
+- "Age" is measured from `completed_at` when set (instances that reached
+  `completed`), otherwise from `created_at`. `updated_at` is intentionally **not**
+  used, because it is user-writable and would let a low-privilege user influence
+  pruning.
+- Pruning runs in a single transaction with foreign-key constraints deferred:
+  matching `df.nodes` rows are deleted first, then the instance rows.
+- Pruning is best-effort: if a pass fails it is logged and retried on the next
+  interval; it never stops workflow execution.
+- If you need to retain terminal history beyond these limits (e.g. for auditing),
+  copy the rows you care about into your own table before they age out.
+
 ---
 
 ## User Isolation & Privileges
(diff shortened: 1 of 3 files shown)


nanookclaw:
Pushed `c455288` to address the PG17 test failure.

The failing helper was reading `pg_catalog.current_database()` as `String`, but PG17 reports that expression as type `name`; the update casts it to `text` before `Spi::get_one::<String>(...)`.

Local verification:
- `cargo fmt --all --check`
- `git diff --check`

`cargo check` is still blocked locally by the same pgrx setup gap (`$PGRX_HOME` missing / `cargo-pgrx` unavailable), so I’m relying on CI for the full PG17 pgrx test rerun.

pinodeca:
Pushed two follow-up commits to the branch.

### `717e021` — fix: correct prune query and harden ranking against forged `updated_at`

- **Security hardening:** the terminal-age ranking used `COALESCE(completed_at, updated_at, created_at)`. Since `completed_at` is only set for `completed` instances (failed/cancelled leave it `NULL`), `terminal_at` for those rows collapsed to `updated_at` — and `updated_at` is user-writable (`GRANT UPDATE (status, updated_at) ON df.instances TO PUBLIC`). A low-privilege tenant could forge a far-future `updated_at` to skew the global keep-window (prematurely deleting other tenants' history) or make their own rows un-prunable. Changed to `COALESCE(completed_at, created_at)` — both are server-controlled and not grantable.
- **Operator-precedence bug:** the age predicate parsed as `(terminal_at < now()) - make_interval(...)`, since explicit `OPERATOR(pg_catalog.*)` operators all share one precedence and are left-associative — yielding `boolean - interval` (`42883`). This was masked in the original PR because the test helper's connection failed before the query ran. Parenthesized `(now() - make_interval(...))`.
- **Test infra:** the unit test built its connection string from the worker-role GUC (defaults to `postgres`, which doesn't exist in the pgrx test cluster → `28000`). Switched to connecting as `current_user`. This is the first unit test that opens a real outbound sqlx connection (all others use in-process SPI). Added a regression case proving a forged future `updated_at` no longer protects/skews a row.

### `a8146d9` — feat: enforce hard 10,000-instance cap on terminal pruning

- Changed the keep predicate from **AND** to **OR**: a terminal instance is now pruned if it falls outside the newest 10,000 (**regardless of age**) **OR** is older than the 30-day retention window. Previously (AND) the retained set could grow unbounded as long as rows were younger than 30 days. Now the retained terminal count can never exceed 10,000.
- Renamed `MIN_KEEP` → `MAX_KEEP` / `min_keep` → `max_keep` to reflect the cap semantics.
- Redesigned the test so a young (3-day-old) row is pruned purely for exceeding the cap, alongside retention-based pruning of old rows and retention of non-terminal (running/pending) rows.
- Documented the policy in `USER_GUIDE.md` (new "Data Retention (Automatic Pruning)" section under Monitoring): hourly best-effort pass, 10,000 hard cap, 30-day window, age basis (`completed_at` else `created_at`; `updated_at` intentionally excluded), terminal-only scope.

**Behavior note:** enforcing the cap removes the prior implicit "keep the newest 10,000 even if older than 30 days" floor — old rows within the top 10,000 are now pruned once they cross the retention window. That's the only way to guarantee "never more than 10,000". If a debugging floor of recent rows is desired even past 30 days, that's a separate two-threshold policy.

Validated: `cargo fmt -p pg_durable -- --check` clean, `cargo clippy --features pg17 --lib` clean, `cargo pgrx test pg17 prune_terminal` passes.

pinodeca:
I force-pushed (with lease) to the PR branch after rebasing on origin/main to pick up a fix for test-flakiness.

Issue #146, opened 2026-05-11
Enabling efficient paginated listing of df.instances for external clients


erdemtuna:
Hi everyone,

We've started to build Durable Pipelines in in the VS Code PostgreSQL extension. We hit a few rough edges in `df.instances` that we'd love your help smoothing. They all share one root cause: external clients (like us) can't paginate `df.instances` efficiently today.

The first durable surface our users will see in the MVP is a paginated **Instances list** sourced straight from `df.instances`. We identified three independent gaps surfaced. They land cleanly in any order, so we've ranked them by impact for convenience, but whichever you pick up first unblocks us.

## Where we got stuck

There's no index on `df.instances.created_at`. The canonical list API `df.list_instances` has no pagination or total count support, and it omits timestamps. The random 8 character `id` can't serve as a proxy for chronological ordering either. So any client that wants a paginated, timestamped Instances list in chronological order has two options: (a) sort the full table on every page request, or (b) fetch a bounded window in one batch and paginate on the client as a workaround.

### Gap 1: `df.instances` has no index for time ordering

Per `src/lib.rs:190, 193`, the only `df.*` indexes today are:

```sql
CREATE INDEX IF NOT EXISTS idx_instances_status ON df.instances(status);
CREATE INDEX IF NOT EXISTS idx_nodes_instance ON df.nodes(instance_id);
```

Notably absent: any index supporting `ORDER BY created_at` or `completed_at`. Every chronological query against `df.instances` requires a full table sort.

We considered using `id` as a proxy for chronological order to lean on the PK index. But `short_id()` (`src/types.rs:91-101`) keeps only the trailing 8 hex characters of a fresh UUIDv4. Fully random, no time component. So `ORDER BY id` is arbitrary, not chronological. No proxy is possible.

### Gap 2: `df.list_instances` lacks pagination and total count

Per `src/monitoring.rs:17-107`, `df.list_instances(status_filter, limit_count)`:

- Runs `SELECT id, label FROM df.instances ORDER BY created_at DESC LIMIT $N` (gated by Gap 1).
- Caps at 10 000 rows.
- Has no `OFFSET` and no total count return value.
- Enriches each row with a `client.get_instance_info(id)` call per row.

There's no canonical API a downstream client can call to render "Page X of Y". 

### Gap 3: `df.list_instances` omits timestamps

[...]

PR #267, opened 2026-06-24
docs: clarify df.status() takes instance_id, not label (#167)

### LINKED ISSUE #167: Clarify that df.status() accepts instance_id, not label
## Summary

`df.status()` accepts an `instance_id`, not a durable function label. This is easy to misuse when callers start workflows with labels and later try to verify the run by passing the label directly to `df.status()`.

## Expected behavior

There should be a clear and ergonomic way to check status for a labeled run. At minimum, docs and examples should show callers how to resolve a label to an `instance_id` before calling `df.status()`.

Possible options:

- Document the pattern of using `df.list_instances()` filtered by label, then passing the selected `instance_id` to `df.status()`.
- Add a helper such as `df.status_by_label(label)` or `df.latest_instance_for_label(label)` if label-based lookup is intended to be a first-class workflow.

## Actual behavior

Passing a label directly to `df.status()` does not return the intended instance status because the argument is interpreted as an `instance_id`.

## Impact

Tests, examples, and clients can appear to verify a labeled workflow while actually querying the wrong identifier.

crprashant:
## What

Documentation-only clarification for `df.status()` (and `df.result()`):

- State explicitly that the argument is an **`instance_id`** returned by `df.start()`, **not** a label. Passing a label returns `NULL`.
- Document the supported **label → `instance_id`** resolution pattern via `df.list_instances()`.
- Fix the `df.status()` example casing in `docs/api-reference.md`: it claimed `'Running'/'Completed'`, but the function reads `df.instances.status` directly and returns the lowercase values `'pending' | 'running' | 'completed' | 'failed' | 'cancelled'`.

Files touched: `docs/api-reference.md`, `USER_GUIDE.md` (Quick Status Check + quick-reference table).

## Why

Closes #167. This is **PR 1** of the agreed, incremental monitoring-functions plan discussed on #167 — the docs-only fix first, with the listing/pagination/label-filter work (touching #87 and #146) to follow in separate small PRs.

## Scope

Docs only — no code, schema, or behavior changes. No first-class label filter is added here; that is a later PR in the plan.

cc @pinodeca @tjgreen42 @AbeOmor

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -183,8 +183,8 @@ df.sql('SELECT 1') ~> df.sql('SELECT 2')
 | `df.break(value)` | Exit loop with **literal** return value (not auto-wrapped as SQL) | `df.break('{"done": true}')` |
 | `df.start(func, label, database)` | Start function (optionally in another database) | `df.start('SELECT 1', 'job')` |
 | `df.cancel(id, reason)` | Cancel function | `df.cancel('a1b2c3d4', 'Done')` |
-| `df.status(id)` | Get status | `df.status('a1b2c3d4')` |
-| `df.result(id)` | Get result | `df.result('a1b2c3d4')` |
+| `df.status(id)` | Get status by instance_id (not label) | `df.status('a1b2c3d4')` |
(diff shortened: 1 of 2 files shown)


pinodeca (approved):
LGTM

PR #275, opened 2026-06-29
Fix N+1 in df.list_instances() with a batched instance-info lookup

crprashant:
## What

Replace the N+1 in `df.list_instances()` with a single batched lookup.

Previously `df.list_instances()`:
1. ran one RLS-filtered SPI query on `df.instances` (newest-first, limited), then
2. built a tokio runtime + duroxide `Client` and called `client.get_instance_info(id)` **once per row** — a separate round-trip per instance — only to read `orchestration_name` (→ `function_name`), `current_execution_id` (→ `execution_count`), and `output`.

Now step 2 is a **single** set-based query over the same worker-credentialed connection. Rather than hand-join duroxide's internal tables, it invokes duroxide-pg's **published** `get_instance_info(TEXT)` SQL function — the same function the duroxide `Client` calls internally — once per id via `CROSS JOIN LATERAL`:

```sql
SELECT gi.instance_id, gi.orchestration_name, gi.current_execution_id, gi.output
FROM unnest($1::text[]) AS t(id)
CROSS JOIN LATERAL <duroxide_schema>.get_instance_info(t.id) AS gi
```

The id set (`$1`) is exactly the RLS-filtered ids from step 1, bound as `text[]`. Depending on the published function's documented OUT columns (rather than duroxide's internal table layout) keeps our coupling on its deliberate contract. `CROSS JOIN LATERAL` returns no row for an id the function yields nothing for, so instances not yet known to duroxide are dropped — preserving the prior per-instance skip-on-missing behavior. Results are reassembled in the original `created_at DESC` order. `status` and `label` continue to come from `df.instances` (authoritative), so all three monitoring APIs still agree on status.

`src/monitoring.rs` (`list_instances` only).

## Why

This is **PR 3** of the agreed, incremental monitoring-functions plan on #167 (issues #87 and #146). Listing N instances previously issued 1 + N queries; for the paginated/external-client listing path (#146) that is the dominant cost. This collapses it to 2 queries total regardless of N.

### Design note (trust boundary)

The batched lookup runs over a **worker-credentialed** connection (`postgres_connection_string()`), the same trust boundary the original per-row `Client` path used — *not* SPI. `df.list_instances()` is `SECURITY INVOKER`, and the duroxide internal schema is not granted to ordinary session roles; routing the lookup through SPI would run it as the calling role and fail with `permission denied` for restricted users (e.g. under the RLS isolation policy). The result set is still constrained to ids already RLS-filtered from `df.instances`, so no cross-user data can leak — the security posture is identical to before.

## Upgrade & compatibility

- **Pure Rust query logic — no SQL/schema contract change.** No upgrade-script edit; `test-upgrade.sh` is not required.
- **Scenario B1:** The new `.so` calls duroxide-pg's published `get_instance_info(TEXT)` over the same worker connection the previous per-row `Client.get_instance_info` path used (its body *is* that lookup), so it works unchanged against all prior `df` schemas.
- No behavior or signature change to `df.list_instances()`: same columns, same `created_at DESC` ordering, same skip-on-missing semantics, same status source.

`0.2.4` is still unreleased, so there is no version bump.

## Testing

Extends `tests/e2e/sql/05_monitoring_and_explain.sql` with a multi-instance case that starts three same-user instances with distinct outputs and asserts (a) they list newest-first (`created_at DESC`) and (b) `function_name`, `execution_count`, and `output` for each agree with `df.instance_info()` — proving the batch lookup reassembles each instance's metadata against the right id rather than scrambling rows. Full suite: 37/37 e2e pass (incl. the RLS-isolation tests that previously caught the SPI trust-boundary regression), `cargo clippy` clean, `cargo fmt` clean.

## Scope

N+1 fix only — no change to the `df.list_instances()` signature, columns, or ordering. `label_filter` + keyset pagination and the truncation policy land in subsequent PRs per the plan.

cc @pinodeca

### DIFF
--- a/src/monitoring.rs
+++ b/src/monitoring.rs
@@ -74,6 +74,26 @@ pub fn list_instances(
         return TableIterator::new(vec![]);
     }
 
+    // function_name, execution_count, and output live in duroxide's internal
+    // schema, which restricted session roles are not granted to read directly.
+    // The previous implementation called the duroxide Client once per instance
+    // (an N+1 of separate provider round-trips). Replace that loop with a single
+    // batched query that invokes duroxide-pg's published get_instance_info(TEXT)
+    // SQL function once per id via LATERAL — the same function the duroxide
+    // Client uses internally — rather than hand-joining duroxide's internal
+    // instances/executions tables. This keeps our dependency on duroxide's
+    // deliberate function contract instead of its internal table layout.
+    // KEEP IN SYNC: we read the function's `instance_id`, `orchestration_name`,
+    // `current_execution_id`, and `output` OUT columns; if duroxide-pg renames
+    // them this query must follow (grep the duroxide-pg migrations for
+    // "get_instance_info"). The lookup runs over a worker-credentialed connection
+    // — the same trust boundary the per-instance Client path used — so it is not
+    // subject to the calling role's privileges on the duroxide schema. The id set
+    // is still constrained to ids already RLS-filtered from df.instances above, so
+    // no cross-user data can leak. Status is taken from df.instances so all
+    // monitoring APIs agree on the status value.
+    let ids: Vec<String> = user_instances.iter().map(|(id, _, _)| id.clone()).collect();
+
     let rt = match tokio::runtime::Builder::new_current_thread()
         .enable_all()
         .build()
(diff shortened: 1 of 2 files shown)


pinodeca (approved):
LGTM

PR #276, opened 2026-06-29
fix: render race branches in explain output

### LINKED ISSUE #274: df.explain() does not render RACE branches
## Summary

`df.explain()` renders a `RACE` node as a single leaf line and drops all of its branches. A race is structurally a parallel fan-out just like `JOIN`, but unlike `JOIN` it has no rendering arm, so its child branches are invisible in the explain tree.

## Reproduction

```sql
-- RACE: branches are dropped
SELECT df.explain( 'SELECT ''winner''' | df.sleep(30) );
```
```
 explain
----------
 RACE
(1 row)
```

Compare with `JOIN`, which renders correctly:

```sql
SELECT df.explain( 'SELECT ''a''' & 'SELECT ''b''' );
```
```
 JOIN (2)
     ║ branch 1:
       ├─SQL: SELECT 'a'
     ║ branch 2:
       └─SQL: SELECT 'b'
```

## Expected vs actual

- **Expected:** the race branches are shown (e.g. `║ branch 1: SELECT 'winner'`, `║ branch 2: SLEEP 30s`), mirroring `JOIN`.
- **Actual:** only `RACE` is printed; both branches are omitted.

## Root cause

`render_children()` in `src/explain.rs` has explicit arms for `LOOP`, `IF`, and `JOIN`, plus a leaf catch-all `_ => {}`. There is **no `RACE` arm**, so a `RACE` node falls into the leaf catch-all and renders zero children.

## Impact

With the new node state-transition model (#263), `df.instance_nodes()` correctly reports a skipped race loser via `inferred_status = 'skipped'`, but `df.explain()` never shows that node at all — so anyone debugging an instance through the explain tree cannot see the abandoned race branch (or its derived status marker). The two views disagree specifically for races.

## Suggested fix

Add a `RACE` arm to `render_children()` mirroring the existing `JOIN` arm (iterate `left_node`/`right_node` and render each as a branch). Verify with both the operator (`|`) and function (`df.race(...)`) forms, and confirm the derived `skipped` marker (`⊘`) appears on the abandoned branch.

## Scope check

- `JOIN` / `join3` — verified working (renders `║ branch N:` for each child); **not** affected.
- `RACE` — affected.

Saithej2k:
## Summary

- Render `RACE` children in `df.explain()` using the same branch layout as `JOIN`
- Show skipped nodes with the `⊘` status marker in explain output
- Add e2e coverage for operator and `df.race(...)` explain output, including the skipped losing branch

Closes #274

## Validation

- `git diff --check`

Not run locally: full e2e suite, which requires the PostgreSQL extension test harness.

### DIFF
--- a/src/explain.rs
+++ b/src/explain.rs
@@ -618,7 +618,7 @@ fn render_children(
                 );
             }
         }
-        "JOIN" => {
+        "JOIN" | "RACE" => {
             let mut branches = Vec::new();
             if let Some(ref left_id) = node.left_node {
                 branches.push(left_id.clone());
@@ -627,13 +627,15 @@ fn render_children(
                 branches.push(right_id.clone());
             }
 
-            // Check for extra nodes in join3
-            if let Some(ref query) = node.query {
-                if let Ok(cfg) = serde_json::from_str::<serde_json::Value>(query) {
-                    if let Some(extras) = cfg["extra_nodes"].as_array() {
-                        for extra in extras {
-                            if let Some(extra_id) = extra.as_str() {
-                                branches.push(extra_id.to_string());
+            if node.node_type == "JOIN" {
+                // Check for extra nodes in join3
+                if let Some(ref query) = node.query {
+                    if let Ok(cfg) = serde_json::from_str::<serde_json::Value>(query) {
+                        if let Some(extras) = cfg["extra_nodes"].as_array() {
+                            for extra in extras {
+                                if let Some(extra_id) = extra.as_str() {
+                                    branches.push(extra_id.to_string());
+                                }
                             }
                         }
                     }
--- a/tests/e2e/sql/05_monitoring_and_explain.sql
+++ b/tests/e2e/sql/05_monitoring_and_explain.sql
@@ -77,6 +77,34 @@ BEGIN
     RAISE NOTICE 'Dry-run explain passed';
 END $body$;
 
+-- Test dry-run explain renders RACE branches for both operator and function forms
+DO $body$
+DECLARE
+    explain_output TEXT;
+BEGIN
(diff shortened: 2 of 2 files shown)


Saithej2k:

pinodeca:
@Saithej2k - thanks for the contribution!

Regarding "Not run locally: full e2e suite, which requires the PostgreSQL extension test harness."

Is there anything I can do to make it easier for contributors like you to do (at least some minimal) local testing?

Note that the devcontainer setup installs pgrx and PG17. Also, if you launch main from a codespace, everything should be ready to go for both testing and launching PG via `scripts/pg-start.sh`

pinodeca:
The change looked good but had merge conflicts (since 263 merged). I was reviewing anyway so I went ahead and rebased then force-pushed (with lease). Tests are running now.

Thanks for allowing maintainers to edit your PR!

UPDATE: I see you force-pushed after me. Thanks for rebasing; your rebased version is better/smaller than mine. Sorry for the noise. I'll merge the PR once CI passes.

pinodeca (approved):
LGTM

Saithej2k:
Thanks for taking a look and for the update. The devcontainer/Codespaces note is helpful. A short docs snippet with the minimal commands for a targeted local check would make this easier next time. I’ll keep an eye on CI.

pinodeca:
@Saithej2k following up on your note about making targeted local testing easier: I opened #277 with a short contributor-docs snippet for minimal local checks in Codespaces/devcontainers and single-test E2E runs. Would you mind taking a look and letting me know if that would have helped for this PR?

Saithej2k:
Thanks for following up. I took a look at #277; yes, that would have helped for this PR. The Codespaces/devcontainer setup note and single-test E2E example make the minimal local check much clearer.

PR #282, opened 2026-07-01
Fail fast in df.start() when the engine hand-off fails

tjgreen42:
## Problem

`df.start()` writes its `df.instances` / `df.nodes` rows in the caller's transaction, then hands the workflow to the durable engine over a separate connection. If that hand-off fails, the error is currently swallowed (logged) and `df.start()` returns normally — so the caller commits an instance row for a workflow that was never started and will never run.

## Change

On a failed hand-off, `df.start()` now raises, aborting the caller's transaction so the instance rows roll back with it. The caller gets a clear, retryable error instead of a silently stuck instance.

The pgrx unit-test build does not run the background worker, so the hand-off always fails there; that build logs instead of aborting (matching the existing `validate_database` test-build carve-out) so `df.start()`'s graph construction stays unit-testable.

## Tests

`tests/e2e/sql/25_start_fail_fast.sql` forces the hand-off to fail (marks the worker not-ready) and asserts `df.start()` raises and commits no row, then confirms a normal start still completes once readiness is restored.

Local: unit 194 · E2E 37 · upgrade 36 · fmt/clippy clean.

### DIFF
--- a/src/dsl.rs
+++ b/src/dsl.rs
@@ -1065,10 +1065,20 @@ pub fn start(
         &instance_id,
         &input_json,
     ) {
-        pgrx::log!(
-            "pg_durable: Warning - failed to start durable function: {}",
-            e
-        );
+        // Fail fast: the durable engine could not accept the start, so abort the
+        // whole df.start() transaction rather than committing an instance row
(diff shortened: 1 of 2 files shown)


pinodeca (approved):
LGTM

PR #283, opened 2026-07-01
Reconcile df control-plane with the durable engine

tjgreen42:
## Problem

`df.start()` writes its `df` control-plane rows in the caller's transaction but hands the workflow to the durable engine over a separate connection. Two gaps followed from that split:

- **Orphaned engine state.** If the caller's transaction rolls back, the `df` rows vanish while the engine keeps an inert record that can never load its (rolled-back) graph and ends up `Failed`. Nothing reclaimed it.
- **Engine state outliving `df` state.** The routine that deleted old terminal instances only removed `df` rows; the matching engine records were never retired, so engine state accumulated without bound.

## Change

The background worker already ran a periodic pass — roughly once an hour — that removed old terminal (`completed`/`failed`/`cancelled`) rows from `df.instances` and `df.nodes` so those control-plane tables don't grow without bound. It only ever touched the `df` tables, never the engine.

This turns that pass into a **reconciliation** of the `df` control-plane and the engine. Each pass:

1. Removes expired terminal instances — their `df.instances`/`df.nodes` rows and their engine records.
2. Reclaims orphaned engine records — the inert, df-less record a rolled-back `df.start()` leaves behind — once it ages past the retention window.

Anything the engine is still tracking with a live `df` row is left untouched.

## Configuration

Two Postmaster-context GUCs govern reconciliation, replacing the hardcoded constants:

| GUC | Default | Meaning |
|-----|---------|---------|
| `pg_durable.retention_days` | `30` | Days a terminal instance and its engine record are retained; also the age bound for reclaiming orphaned engine records. `0` removes terminal instances as soon as the next pass runs. |
| `pg_durable.reconcile_interval` | `3600` | Seconds between reconciliation passes. `0` disables reconciliation. |

A fixed hard cap of 10,000 retained terminal instances still applies. Both settings are documented in `USER_GUIDE.md`.

## Testing

- Unit: expired-instance selection respects age and the hard cap; `select_orphans` excludes df-backed and sub-orchestration ids.
- E2E (`54_reconcile_orphans`): a rolled-back `df.start()` orphan is reclaimed while a live df-backed instance is left intact.

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -1606,36 +1606,65 @@ SELECT started_at, last_seen_at,
 
 The background worker updates `last_seen_at` every ~5 seconds as part of its normal operation.
 
-### Data Retention (Automatic Pruning)
+### Automatic Reconciliation
 
-To keep `df.instances` and `df.nodes` from growing without bound, the background
-worker runs a **best-effort pruning pass roughly once an hour** that deletes old
-**terminal** instances (status `completed`, `failed`, or `cancelled`) and their
-associated `df.nodes` rows. Running and pending instances are **never** pruned,
-regardless of age.
+pg_durable tracks each workflow in its own `df` tables, while the durable engine
+keeps workflow state in a separate schema. To keep those two stores consistent —
+and to keep `df.instances`/`df.nodes` from growing without bound — the background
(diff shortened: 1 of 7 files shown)


pinodeca (approved):
LGTM

PR #284, opened 2026-07-01
De-flake HTTP E2E: replace rate-limited GitHub API call with httpbingo.org

tjgreen42:
## Problem

`tests/e2e/sql/06_http_and_ssrf.sql`'s "JSON API with vars and loop" case fetched `https://api.github.com/repos/microsoft/duroxide/commits`. GitHub's API rate limit makes that request fail intermittently in CI (`TEST FAILED: GitHub API fetch status = failed`), flaking the whole E2E job.

## Change

Point the case at `https://httpbingo.org/get` — the same non-rate-limited HTTP test endpoint the rest of this file already uses — so it is deterministic. Both hosts are already in the `http-allow-test-domains` allowlist, so no code change is needed.

## Test

`./scripts/test-e2e-local.sh 06_http_and_ssrf` passes.

### DIFF
--- a/tests/e2e/sql/06_http_and_ssrf.sql
+++ b/tests/e2e/sql/06_http_and_ssrf.sql
@@ -263,85 +263,84 @@ END $$;
 DROP TABLE _test_http_vars;
 SELECT df.clearvars();
 
--- === Test: 19_github_api ===
+-- === Test: 19_http_json_api_loop ===
 
-DROP TABLE IF EXISTS github_commits;
-CREATE TABLE github_commits (
+DROP TABLE IF EXISTS http_loop_results;
+CREATE TABLE http_loop_results (
     id SERIAL PRIMARY KEY,
-    sha TEXT UNIQUE,
-    author TEXT,
-    message TEXT,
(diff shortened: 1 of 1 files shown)


pinodeca (approved):
From Opus 4.8:

### Minor nit (non-blocking)
`httpbingo.org` runs `go-httpbin`, which returns `args` and `headers` as **arrays of strings** (multi-valued), unlike the original `httpbin.org` which returns scalars. So:

```sql
body->'args'->>'source'          -- stores  ["pg_durable"]      (not  pg_durable)
body->'headers'->>'User-Agent'   -- stores  ["pg_durable-test"] (not  pg_durable-test)
```

The test still passes because it only asserts `result_count > 0`, and `url` (the `ON CONFLICT` key) is a scalar. But the `source` / `user_agent` columns—and the final display `SELECT`—will contain JSON-array-serialized text rather than clean scalars. If you want the displayed values to look right, index the first element:

```sql
body->'args'->'source'->>0,
body->'headers'->'User-Agent'->>0
```

PR #302, opened 2026-07-27
Support http multipart/form-data uploads

sweatybridge:
Adds a dedicated SQL function `df.http_multipart(url, method, parts, headers, timeout)` for multipart/form-data requests (file uploads, form posts). Binary part payloads are base64-inlined in the config JSON (which is durably checkpointed).

Briefly verified e2e but lacks polish. Would appreciate a thorough review and any edits to get it production ready.

Thank you!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1609,6 +1609,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "mime"
+version = "0.3.17"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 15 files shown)


sweatybridge:

pinodeca (approved):
Thanks for this contribution @sweatybridge.

I'm going to take this change as-is since it's passing CI gates. Then I will follow-up with one or two PRs that clean up some issues I found and provide a real-world example.

PR #331, opened 2026-08-04
Fix deep workflow graph composition

### LINKED ISSUE #327: df.seq silently emits its own serialized expression tree as SQL beyond ~100 chained steps
**Component:** DSL graph construction (`df.seq` / `~>`)
**Versions:** pg_durable 0.2.1, duroxide 0.1.28, PostgreSQL 17.9 (Azure HorizonDB)

## Summary

Building a sequential chain of more than roughly 100 steps does not fail at
construction time. `df.start` returns an instance ID normally, and the instance
is then guaranteed to fail at execution with:

```
SQL execution failed: error returned from database: syntax error at or near "{"
```

The cause is visible in the stored graph: past a certain point `df.seq` stops
materializing node handles and instead writes its own serialized expression tree
into `df.nodes.query`, where the worker subsequently executes it as SQL.
Inspecting `df.nodes.query` for the failing node shows literal JSON:

```
{"node_type":"THEN","left_node":{ …
```

This is a correctness bug rather than a capacity limit. A chain that exceeds the
builder's limit produces a *corrupt* graph, not a rejected one, and the corruption
is only discovered after the instance has been accepted and started.

## Reproduction

```sql
-- h := df.sql(step_1); then repeatedly:
select df.seq(<h>, <next step expr>);   -- -> h
select df.start(<h>, label => <label>);
```

Repeat the `df.seq` fold N times.

| N | Outcome |
|---|---|
| 50 | completes |
| 100 | completes |
| 200 | fails: `syntax error at or near "{"` |
| 1000 | fails: `syntax error at or near "{"` |

Per-step cost up to the boundary is 102–110 ms per checkpointed step, so the
failure is not a timeout.

## Node materialization at the break

Counting rows in `df.nodes` for the failing instances shows construction stops
well short of the requested depth:

| Requested depth | SQL nodes | THEN nodes |
|---|---|---|
| 200 | 73 | 72 |
| 1000 | 111 | 110 |

The counts are not a clean function of the requested depth, which suggests the
boundary is a size/recursion limit inside the builder rather than a step count.

## Impact

Any workload that constructs a long sequential pipeline — a per-row ETL fold, a
generated migration chain, a loop unrolled at build time — will be accepted,
persisted, started, and then fail. Because the failure surfaces only at execution,
callers that check `df.start` for success will believe the workflow was durably
enqueued when the stored graph cannot run.

## Suggested handling

Fail loudly at the construction boundary. Raising an error from `df.seq` when the
tree can no longer be materialized would be strictly better than emitting a graph
that is guaranteed to fail, and would let callers chunk or restructure before
anything is persisted. If a legitimate depth ceiling exists, documenting it and
enforcing it explicitly would also resolve this.

## Provenance

Found during a stress campaign against pg_durable 0.2.1 on Azure HorizonDB
(PostgreSQL 17.9), 3 August 2026. Phases covered submission storms, chain depth,
fan-out, signals, cancellation, and durable-state growth. The full report is
internal to Microsoft; the reproduction above is self-contained.

pinodeca:
## Summary

- store `Durofut` children as opaque `RawValue` JSON objects so each graph level deserializes independently
- parse children on walker descent and validate graph depth before `df.explain()` traversal
- preserve PostgreSQL stack/resource errors in `df.ensure_durofut()` and update the 0.2.6 upgrade path
- add wire-format, composer, depth-boundary, operator, and 200-step execution regressions

Supersedes #328.
Fixes #327.

## Validation

- `cargo build --features pg17`
- `./scripts/test-unit.sh` (263 passed, 16 ignored)
- `./scripts/test-e2e-local.sh 09_graph_and_validation`
- `./scripts/test-upgrade.sh` (66 passed)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ Pre-1.0 note: while `pg_durable` is in major version `0`, minor releases may inc
 
 ## [0.2.6] - Unreleased
 
+### Fixed
+
+- **Deep workflow composition (#327):** workflow graphs deeper than serde_json's 127-level recursion limit no longer silently collapse into SQL text. Nested children are deserialized one graph level at a time, and `df.explain()` now enforces the configured graph-depth limit before traversal.
+
 ## [0.2.5] - 2026-07-30
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ pg_test = []
 [dependencies]
 pgrx = "=0.16.1"
 serde = { version = "1.0", features = ["derive"] }
-serde_json = { version = "1.0", features = ["preserve_order"] }
+serde_json = { version = "1.0", features = ["preserve_order", "raw_value"] }
 uuid = { version = "1.0", features = ["v4", "serde"] }
 percent-encoding = "2.3"
 
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -309,6 +309,12 @@ SELECT df.start(
 );
 ```
 
+### Graph Depth
+
+A workflow graph can contain at most 256 nested composition levels. `df.start()` and
+`df.explain()` reject deeper graphs with a maximum-nesting-depth error. Split a larger workflow
+into smaller sub-workflows when it would exceed this limit.
+
 ### Variable Substitution
 
 Use `$name` to reference named results in subsequent steps:
--- a/docs/upgrade-testing.md
+++ b/docs/upgrade-testing.md
@@ -203,6 +203,14 @@ gate, so they never need to be added to the exclude list.
 Each schema-changing PR should add a section here documenting what changed,
 what the upgrade script handles, and any backward compatibility considerations.
 
+### v0.2.5 → v0.2.6
+
+#### Preserve parser resource errors in `df.ensure_durofut()`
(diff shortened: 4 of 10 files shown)


pinodeca:
Thanks for the detailed follow-through. The remaining silent-corruption items you flagged are addressed in a narrow follow-up: #328.

**Carried over from your review:**
- `Durofut::ensure()` no longer swallows an unparseable envelope into a raw SQL node. It now fails loudly, keying off the parse `Err` plus a generic-JSON `node_type` probe — robust to serialized field order, so it avoids the `looks_like_serialized_durofut` prefix sniff you (rightly) called out. This also surfaces the new `deserialize_raw_object` "Durofut children must be JSON objects" rejection instead of silently wrapping it.
- Review point 1: `collect_nodes()` in `df.explain()` now raises `pgrx::error!` instead of `.expect()`-panicking, matching `insert_nodes()` in `df.start()`.

**Deliberately left as-is:**
- `is_durofut()` (`unwrap_or(false)`) — returning `false` is the correct contract for a boolean predicate.
- The `for_each_config_child` / `transform_config_children` top-level `Err(_)` guards already fail loudly on malformed `condition_node` / `extra_nodes`; the top-level arm only skips a genuinely non-JSON `query` field.
- Review point 2 (depth errors surfacing at `df.start()`/`df.explain()`) — treated as the intended trade-off you described, not a bug.

**Tests:** loud-failure on a corrupt envelope via `df.seq`, envelope detection independent of field order, and a deep JOIN chain to lock in that JOIN/RACE share the corrected opaque-child path. I stopped short of deep RACE/`loop`/`as` chains and an E2E test — happy to add those if you'd like them pinned explicitly.

PR #335, opened 2026-08-05
Share iterative graph materialization

pinodeca:
## Summary

- add a shared iterative `flatten_graph` pass for validation and materialization
- use the same flat `FunctionNode` representation in `df.start()` and `df.explain()`
- report graph errors with child paths and preserve the existing `df.nodes.query` format
- document pre-order ID assignment and its deferred-foreign-key requirement

## Validation

- `cargo test --features pg17 types::tests --lib` (69 passed)
- `./scripts/test-unit.sh` (274 passed, 16 ignored)
- `./scripts/test-e2e-local.sh 09_graph_and_validation`
- `./scripts/test-upgrade.sh` (70 passed across supported schemas 0.2.2 through 0.2.5)
- `cargo clippy --features pg17 --lib -- -D warnings`
- `cargo fmt -p pg_durable -- --check`
- `cargo check --features pg17`

PR #337, opened 2026-08-05
Batch graph node inserts

pinodeca:
## Summary

- insert flattened graph nodes in bounded 1,000-row multi-value statements
- remove obsolete per-node conflict handling now that graph IDs are assigned uniquely in memory
- preserve both current and legacy `login_role` column layouts
- add SQL-shape unit tests and a 1,002-node transactional boundary test

## Validation

- `./scripts/test-unit.sh` (283 passed, 16 ignored)
- `./scripts/test-e2e-local.sh 09_graph_and_validation`
- `./scripts/test-upgrade.sh` (all 70 checks passed across supported schemas 0.2.2 through 0.2.5)


pinodeca:
## Performance results

I compared #337 with #335 using preconstructed shallow-wide graphs, so graph composition was excluded. Each sample called `df.start()` inside a transaction that was rolled back, measuring parsing, validation, flattening, instance reservation, and node persistence without allowing rows or workflow execution to accumulate. Medians below are from 15 retained samples after 5 warmups.

| Nodes | #335 median | #337 median | Improvement |
|---:|---:|---:|---:|
| 100 | 30.937 ms | 9.269 ms | 3.34x |
| 500 | 137.659 ms | 32.270 ms | 4.27x |
| 1,000 | 272.070 ms | 61.392 ms | 4.43x |
| 5,000 | 1,347.918 ms | 256.265 ms | 5.26x |

`pg_stat_statements` confirmed the structural change: #335 issued one `df.nodes` INSERT per node, while #337 issued one INSERT per 1,000-node chunk. Boundary checks produced exactly 1, 1, and 2 INSERTs for 999, 1,000, and 1,001 nodes, and 5 INSERTs for 5,000 nodes.

The measured result matches the intended optimization: #337 removes repeated per-node SPI planning/execution and substantially reduces `df.start()` latency, with the benefit increasing as graph size grows.

PR #209, opened 2026-06-05
Upgrade `pg_durable` to pgrx v0.18.1

eeeebbbbrrrr:
This PR upgrades the project to pgrx v0.18.1 in 7e7d0c2d443f376788a2bb812c280fd619fb2866 and then fixes some issues running the test suite on macOS in e275752eb328f0c42761eb310300c14f37b66cfa.

As evidenced by the commit attributions, this was 100% written by Claude.  

Nonetheless, pgrx v0.18.1 is a **significant** upgrade over 0.16.1 as it completely eliminates a full round of compilation on every build, which ought to make CI and your developers of this project a lot happier.

I'm happy to participate in the review of this and cleanup anything y'all think is necessary.

The test suite passed locally for me on macOS (with e275752eb328f0c42761eb310300c14f37b66cfa) for pg17 and pg18.  I did not run the Docker e2e tests as I do not (and will not ever) have Docker installed.

PR #356, opened 2026-08-24
Update pgrx to 0.19.2

thomcc:
I'm not actually sure this is worth doing.

Half of the reason for this PR is checking if everything was working with my dev/PR/etc setup, and the other half was to see if it was easy (since it is always possible for the need to perform such an update to occur suddenly, and it would be useful to have a branch for it in the back pocket if such things happen).

I am interested in seeing if CI passes, however, so I'll at least check that first.

### DIFF
--- a/.devcontainer/onCreateCommand.sh
+++ b/.devcontainer/onCreateCommand.sh
@@ -43,11 +43,11 @@ else
 fi
 
 # Install cargo-pgrx
-echo "Installing cargo-pgrx 0.16.1..."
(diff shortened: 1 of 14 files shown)


thomcc:
I assume this isn't worth doing for the same reason as https://github.com/microsoft/pg_durable/pull/209. (But it was still worth it to get everything set up for me)

thomcc-work:
Closing, the task of eventually getting us up to date is tracked here: https://github.com/microsoft/pg_durable/issues/391

PR #361, opened 2026-08-27
Improve secrets handling

thomcc:
This is still very WIP. Several parts of it (script changes) were made just to get things working in my env (I'll pull them into a separate PR later, or adjust my env so that they aren't needed), and it only handles redaction and stripping of logs. Still working on this.


thomcc:
I've moved everything over to a dedicated work account (@thomcc-work), for various reasons. Going to reopen under that.

PR #364, opened 2026-08-28
Use raw role names verbatim for worker connections

pinodeca:
## Summary

- pass the exact catalog role name to the PostgreSQL connection builder
- remove SQL-identifier normalization from the wire-protocol username path
- preserve current host and application-name connection behavior
- add negative coverage for quote-wrapped role names that resemble a superuser
- add positive coverage proving a literal quote-wrapped role executes as itself
- update the unreleased changelog

## Security impact

Catalog role names are literal strings, not quoted SQL identifiers. Reinterpreting a quote-wrapped name could collapse it onto a distinct login role and make the worker execute under the wrong identity.

## Validation

- `cargo check --features pg17`
- `./scripts/test-e2e-local.sh --verbose 17_superuser_guc` (passed)
- `./scripts/test-e2e-local.sh --verbose 49_quoted_role_names` (passed)

### DIFF
--- a/tests/e2e/sql/17_superuser_guc.sql
+++ b/tests/e2e/sql/17_superuser_guc.sql
@@ -14,6 +14,7 @@
 --   3. Forgery caught by load_function_graph (instance-level rejection).
 --   4. Forgery caught by execute_sql (node-level rejection, post-cache tamper).
 --   5. Cross-iteration forgery neutralized by the frozen graph snapshot.
+--   6. A literal quoted catalog role cannot authenticate as postgres.
 --
 -- "GUC on + superuser succeeds" is implicitly covered by every other E2E test
 -- (standard phase runs as postgres with enable_superuser_instances = on).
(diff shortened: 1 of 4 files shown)


tests/e2e/sql/17_superuser_guc.sql
```diff
@@ -341,6 +342,80 @@ END $$;
...
+        SELECT oid::regrole FROM pg_roles WHERE rolname = chr(34) || 'postgres' || chr(34)
+    )
+    WHERE id = inst;
+    UPDATE df.nodes
+    SET submitted_by = (
+        SELECT oid::regrole FROM pg_roles WHERE rolname = chr(34) || 'postgres' || chr(34)
+    )
+    WHERE instance_id = inst;
+END $forge_quoted_role$;
+RESET SESSION AUTHORIZATION;
```
pinodeca, line 394: **P2 — Make submission and forgery atomic.** `df.start()` commits before these updates, so the worker can load or execute the graph as the original `df_e2e_user` first. That role also cannot insert into `su_guc_quoted_role_sentinel`, producing exactly the asserted `failed` status and empty sentinel; the test can therefore pass without exercising the quoted-role connection path. Please wrap submission plus both identity updates in one explicit transaction so the worker cannot observe the instance until the forged identity is committed. It would also help to assert the failure detail identifies authentication of the literal quoted role rather than accepting any failure.

pinodeca (commented):
One test reliability issue found. The production change is directionally correct and the positive `current_user` assertion covers the central quote-wrapped-role regression, but the new negative superuser test has a race that can yield a false pass.

PR #377, opened 2026-09-04
Add failure-isolated loop iterations

tjgreen42:
Long-running scheduled workflows should survive an isolated iteration failure instead of terminating permanently. This is motivated by the pg_textsearch compaction backstop in timescale/pg_textsearch#476, where one failed compaction must not prevent later scheduled attempts.

> **Experimental:** The `continue_on_failure` syntax is subject to change in future releases.

## Summary

- add the unified `df.loop(body, condition DEFAULT NULL, continue_on_failure DEFAULT false)` API for infinite and conditional loops
- run each opted-in body iteration as a child orchestration; consume errors returned by body SQL, HTTP, and multipart activities, skip the condition for that iteration, and continue from the top
- merge named results after successful bodies so the condition can use them; keep condition failures, including nested-loop conditions, fatal
- keep malformed graph/protocol data, child-ID collisions, and orchestration/runtime failures fatal
- deterministically prioritize fatal `JOIN` sibling failures over `df.break()` and recoverable activity failures inside failure-isolated iterations, while preserving historical first-error behavior elsewhere
- raise the loop iteration backstop from 100,000 to 8,388,608 (`2^23`), approximately 80 years at five-minute intervals
- update the SQL upgrade path, documentation, examples, architecture description, and pg_durable SQL skill

## Upgrade note

Histories below iteration 100,000 replay unchanged. A loop that already recorded the old terminal-failure path at that boundary cannot replay under the new binary; drain such long-running loops before upgrading when continuity is required.

## Testing

- 327 unit tests passed
- 12 loop-focused E2E tests passed
- all 109 upgrade tests passed

### DIFF
--- a/.agents/skills/pg-durable-sql/SKILL.md
+++ b/.agents/skills/pg-durable-sql/SKILL.md
@@ -93,9 +93,16 @@ df.if(condition TEXT, then_branch TEXT, else_branch TEXT) → TEXT
 -- result_name is a capture from |=> earlier in the graph.
 df.if_rows(result_name TEXT, then_branch TEXT, else_branch TEXT) → TEXT
 
--- Loop — infinite or while-condition
-df.loop(body TEXT) → TEXT                            -- Infinite loop
-df.loop(body TEXT, condition TEXT) → TEXT             -- While-loop: repeats while condition is truthy
+-- Loop — one unified signature
+df.loop(
+    body TEXT,
+    condition TEXT DEFAULT NULL,
+    continue_on_failure BOOLEAN DEFAULT false
+) → TEXT
+-- NULL condition: infinite loop.
+-- Non-NULL condition: do-while semantics; evaluate it after each successful body.
+-- With continue_on_failure => true, a consumed body activity failure skips the
+-- condition and starts the next iteration. Condition and non-application failures are fatal.
 
 -- Break from enclosing loop
 df.break() → TEXT                                    -- Exit with NULL
@@ -227,7 +234,7 @@ Automatically available during execution:
 
 ## Condition Evaluation (Truthiness)
 
-Used by: `?>`, `!>`, `df.if()`, `df.loop(body, condition)`
+Used by: `?>`, `!>`, `df.if()`, and the optional condition in `df.loop()`
 
 The first column of the first row is evaluated:
 
@@ -345,7 +352,7 @@ SELECT df.start(
 -- Cancel with: SELECT df.cancel('instance_id', 'Stopping heartbeat');
 ```
 
-### While-Loop with Break
+### Loop with Break
 
 ```sql
 SELECT df.start(
@@ -360,6 +367,26 @@ SELECT df.start(
     'counted-loop'
 );
 ```
+
+### Conditional Loop with Failure Continuation
+
+```sql
+SELECT df.start(
+    df.loop(
+        'SELECT process_next_item()',
+        'SELECT EXISTS (
+            SELECT 1 FROM work_queue WHERE status = ''pending''
+        )',
+        continue_on_failure => true
+    ),
+    'resilient-worker'
+);
+```
+
+This is a do-while loop: after a successful body execution, the condition is
+evaluated and the loop continues while it is truthy. After a consumed body
+activity failure, the condition is skipped and the body starts again.
+Condition failures and non-application failures remain fatal.
 
 ### Cron Scheduled Job
 
--- a/.agents/skills/pg-durable-sql/SKILL.md
+++ b/.agents/skills/pg-durable-sql/SKILL.md
@@ -102,7 +102,8 @@ df.loop(
 -- NULL condition: infinite loop.
 -- Non-NULL condition: do-while semantics; evaluate it after each successful body.
 -- With continue_on_failure => true, a consumed body activity failure skips the
--- condition and starts the next iteration. Condition and non-application failures are fatal.
+-- condition and starts the next iteration. The loop's condition and
+-- orchestration/runtime failures remain fatal.
 
 -- Break from enclosing loop
 df.break() → TEXT                                    -- Exit with NULL
@@ -386,7 +387,8 @@ SELECT df.start(
 This is a do-while loop: after a successful body execution, the condition is
 evaluated and the loop continues while it is truthy. After a consumed body
 activity failure, the condition is skipped and the body starts again.
-Condition failures and non-application failures remain fatal.
+Errors returned by body SQL, HTTP, and multipart activities are consumable.
+Condition and orchestration/runtime failures remain fatal.
 
 ### Cron Scheduled Job
 
--- a/.agents/skills/pg-durable-sql/SKILL.md
+++ b/.agents/skills/pg-durable-sql/SKILL.md
@@ -99,6 +99,7 @@ df.loop(
     condition TEXT DEFAULT NULL,
     continue_on_failure BOOLEAN DEFAULT false
 ) → TEXT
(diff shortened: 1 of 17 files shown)


src/dsl.rs
```diff
@@ -321,21 +322,48 @@ pub fn wait_for_schedule(cron_expr: &str) -> String {
...
         left_node: Some(body_fut.into_raw()),
         condition_node,
+        query,
         ..Default::default()
     }
     .to_json()
 }
 
+#[pg_extern(name = "_loop_legacy", schema = "df")]
+pub fn loop_fn(body: &str, condition: default!(Option<&str>, "NULL")) -> String {
```
thomcc-work, line 354: This is (along with the SQL) is a clever way to avoid breaking ABI when changing the signature of a Rust function. We should consider documenting it somewhere (possibly in the copilot instructions, or something that it links to). Not saying to do it as part of this PR, though.

.agents/skills/pg-durable-sql/SKILL.md
```diff
@@ -93,9 +93,16 @@ df.if(condition TEXT, then_branch TEXT, else_branch TEXT) → TEXT
...
 
--- Loop — infinite or while-condition
-df.loop(body TEXT) → TEXT                            -- Infinite loop
-df.loop(body TEXT, condition TEXT) → TEXT             -- While-loop: repeats while condition is truthy
+-- Loop — one unified signature
+df.loop(
+    body TEXT,
+    condition TEXT DEFAULT NULL,
+    continue_on_failure BOOLEAN DEFAULT false
+) → TEXT
```
thomcc-work, line 101: It was already there but the fact that this uses a Unicode arrow instead of `->` (since it's nominally a SQL code block) kind of bothers me. To be clear, I'm not asking you to change this here, it's just odd.

src/orchestrations/execute_function_graph.rs
```diff
@@ -930,8 +1019,8 @@ const LOOP_MIN_ITER_DURATION: Duration = Duration::from_secs(1);
 
 /// Maximum loop iterations before the orchestration is forcibly terminated.
 /// This prevents runaway infinite loops from consuming resources indefinitely.
-/// At the minimum 1-second rate limit, this allows ~27 hours of looping.
-const MAX_LOOP_ITERATIONS: u64 = 100_000;
+/// At the minimum 1-second rate limit, this allows over 97 days of looping.
+const MAX_LOOP_ITERATIONS: u64 = 1 << 23;
```
thomcc-work, line 1053: After my manual review I threw it into the AI to see if it said anything. It complained about this change:

> There is a **narrow binary-upgrade window**: an old execution has reached the 100,000-iteration limit and scheduled its failed-node-status activity, but has not finished terminating. Replaying that history with the new binary no longer takes the limit-failure branch. Instead, it schedules the next `utc_now()` activity where history expects `update_node_status(..., "failed")`.
>
> Duroxide rejects that activity mismatch. If root-instance finalization has not happened, the workflow can remain reported as `running` rather than failing normally. This does not affect ordinary earlier iterations or already-terminal executions.
>
> **Suggested correction:** Preserve the old cap for legacy orchestration inputs and persist/version the higher cap for new workflows—or defer this independent cap increase until its replay compatibility is handled.

I'm not sure I fully understand the details (I think it's basically replaying things that failed under the old cap won't fail now that the cap has changed, which will be an difference which will upset duroxide? I suppose that makes sense).

src/orchestrations/execute_function_graph.rs
```diff
@@ -2037,6 +2189,147 @@ mod tests {
         assert!(graph_retry_budget_exceeded(u32::MAX));
     }
 
+    #[test]
+    fn loop_iteration_backstop_is_two_to_the_twenty_third() {
```
thomcc-work: This is an odd test. I'm not sure it's actually useful to assert the value of a constant from the test. If we have a hard limit that it must be a power of two, IMO that check would probably be better as a compile time assertion, e.g. `const _: () = assert!(MAX_LOOP_ITERATIONS.is_power_of_two());`, probably outside the tests and back near the constant itself, so someone notices it. But if we don't need this, I'd ditch it entirely.

src/orchestrations/execute_function_graph.rs
```diff
@@ -1556,25 +1724,17 @@ async fn execute_join_node(
...
-                    )
-                    .await;
+                    return fail_loop_child_future(ctx, graph, &branch_ids[i], error).await;
                 }
-                return Err(NodeError::Failure(format!(
-                    "JOIN branch {} failed: {}",
-                    i + 1,
-                    e
-                )));
+                return Err(error);
```
thomcc-work: More thoughts from the robots. It seems to think that retrying a recoverable error from a JOIN branch in a df.loop might hide a fatal error that occurs in a sibling. That also makes sense to me, assuming I'm interpreting it right. The actual words of the robot follow: 

> In an opted-in loop, suppose the left JOIN branch fails an SQL activity, while the right branch hits a non-recoverable error—for example, strict variable substitution rejects a captured SQL NULL. JOIN waits for both branches, but returns the left branch’s recoverable `Application` error without inspecting the right branch’s fatal error. The loop consequently continues, contrary to the intended distinction between activity failures and malformed data/runtime faults.
>
> The first-error selection predates this PR, but previously it always terminated the loop. Making that error recoverable newly allows fatal sibling failures to be swallowed.
>
> **Suggested correction:** Inspect all settled JOIN outcomes before returning a recoverable error, and deterministically prioritize non-recoverable failures. Preserve existing replay behavior for legacy workflows.

thomcc-work (commented):
I'm still not at the point of pg_durable understanding where I can confidently comment as to whether the change makes sense to do it this way. Really, the only thing I can comment on is general style/approach in the abstract. Because of this, I'm just going to submit this review as "comment" rather than "approving" but it seems mostly viable (assuming you look at the genuine problems and either address them, or don't think we need to).

I am glad I took a look though, since I hadn't considered handling signature changes that way. It's still not practical for cases like mine, (`df.http` is GRANT city, and it's apparently hard to migrate those properly), but it's a really good technique to know.

I also asked a robot to take a look at it on my end, and it surfaced two things, both of which *seem* to make sense to me, but I couldn't say for sure.

.agents/skills/pg-durable-sql/SKILL.md
tjgreen42: It is kind of odd :)  File under "em-dash".

src/orchestrations/execute_function_graph.rs
tjgreen42: I think this is unlikely to be a problem in practice.  The 100K iteration limit is just a guardrail that is supposed to be beyond what anyone means to execute.  But it wasn't quite enough breathing room for a cron-type job that is supposed to run essentially indefinitely.  Plus, we are still officially pre-GA, so minor hiccups are okay.

src/orchestrations/execute_function_graph.rs
tjgreen42: Yeah, I agree with you, no human would write a test like this.  I'll remove it.

src/orchestrations/execute_function_graph.rs
tjgreen42: Hmm, this looks worth fixing.

src/dsl.rs
tjgreen42: Copilot came up with this on its own, so I think we can probably skip the note.

pinodeca (approved):

pinodeca (approved):

PR #378, opened 2026-09-04
Improve secrets handling

thomcc-work:
Updated version of https://github.com/microsoft/pg_durable/pull/361, now on the correct account.

Does 3 things:

- Adds `pg_durable.log_workflow_sql` postmaster GUC. This controls whether or not the worker will write the fully substituted SQL text of a workflow node to the server log.
- Redacts URLs in logs and errors. Query parameter values, user info, and the fragment are elided. This means that an Azure SAS token, or `?api-key=` won't reach the PG server log, `df.nodes.error`, or durable execution history. Other parts of the URL are preserved. Error text from the HTTP client is scrubbed too, since it interpolates the URL.
- Adds documentation that's honest about `df.vars` and potential exposure to USER_GUIDE.md.

Blocks most of the other work I have for improving `df.http`.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -588,6 +588,11 @@ Sets a workflow variable for the current user (before `df.start()`). Each user h
 SELECT df.setvar('api_url', 'https://api.example.com');
 ```
 
+> **Not for credentials.** Values are stored as plaintext in `df.vars`, and `df.start()` copies
+> every variable you own into durable execution history. Using `{varname}` keeps the value out of
+> `df.nodes.query` but not out of history. See
+> [Variables and secrets](../USER_GUIDE.md#variables-and-secrets).
+
 ---
 
 ### df.getvar(name)
@@ -684,7 +689,7 @@ SELECT df.revoke_usage('app_role');
 
 ## Server Configuration (GUCs)
 
-These settings are configured via `ALTER SYSTEM SET` or `postgresql.conf` and take effect after `SELECT pg_reload_conf()` (no restart required).
+These settings are configured via `ALTER SYSTEM SET` or `postgresql.conf`. When they take effect depends on the context listed for each one: `SUSET` settings apply after `SELECT pg_reload_conf()`, while `POSTMASTER` settings require a server restart. GUCs read by the background worker are `POSTMASTER`, because the worker does not process a configuration reload.
 
 ---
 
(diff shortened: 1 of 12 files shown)


thomcc-work:
Hm, needing to sign the CLA is a bit odd (I would have thought that was automatic given that I'm an employee and am part of the Microsoft org), but whatever. I would have signed the CLA on my old account prior to joining MS, so maybe it's just always needed.

@microsoft-github-policy-service agree company="Microsoft"

docs/api-reference.md
```diff
@@ -684,7 +689,7 @@ SELECT df.revoke_usage('app_role');
 
 ## Server Configuration (GUCs)
 
-These settings are configured via `ALTER SYSTEM SET` or `postgresql.conf` and take effect after `SELECT pg_reload_conf()` (no restart required).
+These settings are configured via `ALTER SYSTEM SET` or `postgresql.conf`. Each one lists its context: `SUSET` settings take effect after `SELECT pg_reload_conf()`, while `POSTMASTER` settings require a restart. Every GUC read by the background worker is `POSTMASTER`, because the worker does not process a configuration reload.
```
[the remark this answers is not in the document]
thomcc-work: Done.

src/redact.rs
```diff
@@ -0,0 +1,261 @@
...
+/// lone token is indistinguishable from a name.
+fn redact_query(query: &str) -> String {
+    let mut out = String::with_capacity(query.len());
+
+    for (i, pair) in query.split('&').enumerate() {
+        if i > 0 {
+            out.push('&');
+        }
+        match pair.split_once('=') {
+            Some((_, "")) => out.push_str(pair),
```
pinodeca: Non-blocking P2: please consider redacting empty-valued pairs too. A raw opaque token with base64-style padding, such as `?c2VjcmV0=`, lands in this branch and is copied verbatim, even though the adjacent bare-token case fails closed. Since `name=` is ambiguous between an ordinary empty parameter and an opaque padded token, the safer behavior is to redact the whole pair (with a regression test covering padded tokens).

pinodeca (approved):
Approved. The URL scrubbing and SQL-log control look sound overall, and the focused redaction tests pass.

I left one non-blocking P2 inline for consideration: empty-valued query pairs are preserved verbatim, so an opaque base64-style token ending in `=` can escape redaction. I do not think this needs to block the PR, but failing closed on that ambiguous form would better match the helper's security posture.

Other reviewer findings, recorded here for awareness and possible follow-up:

- **DNS SSRF audit classification:** correctness and reliability reviewers independently noted that `is_ssrf_block_error(e.to_string())` may not see resolver errors nested in reqwest's source chain. The request remains blocked, and this appears to predate this PR, so I filtered it as a follow-up diagnostics/auditing issue rather than a regression.
- **`log_workflow_sql` defaults to `on`:** a security reviewer preferred a secure-by-default `off`. The current default is an intentional, documented compatibility choice, so I did not treat it as a defect; it may still merit a separate product/security decision.
- **5xx response previews and final workflow results can expose secrets:** this is pre-existing and explicitly documented as outside the URL-redaction guarantee, so it is not a blocker for this scoped change.
- **Response-body read diagnostics:** removing the attached URL can leave generic errors without the redacted endpoint/status context. This is an observability tradeoff, not a demonstrated correctness or security regression; structured safe context could be added later.
- **Test coverage:** there is no end-to-end server-log assertion proving `log_workflow_sql=off` suppresses both activity and SQLx statement text, nor an E2E assertion that credential-bearing HTTP/multipart failures stay out of persisted errors. Unit coverage and CI are green, so I treated these as follow-up hardening rather than blockers.
- **Reqwest-owned error wording in tests:** a testing reviewer flagged remaining dependency-message assertions as brittle. This is low severity and does not affect runtime behavior.

src/redact.rs
thomcc-work: Fixed.

pinodeca (approved):
Thanks for addressing the P2 issue from my previous review. Re-approved.

PR #379, opened 2026-09-04
HTTP client caching in the activity

thomcc-work:
This reuses the client if in the same process, which can speed things up a good amount if requests are made to the same hosts (among other cases). It has a drawback around only reading env vars when the client is created, so proxy settings wouldn't update as readily. I don't think we care about this, but if we do, we could probably detect it and handle it (it would just be annoying).

Timeout is now configured on a per-request rather than per-client basis. There's some testing for this.

Note that this PR applies on top of #378, and so it has the (unrelated) changes from there in it.

### DIFF
--- a/docs/http-security.md
+++ b/docs/http-security.md
@@ -344,6 +344,15 @@ inherit another's authentication by landing on a warm connection. The
 per-node timeout is applied to each request rather than to the client for the
 same reason — it is caller configuration, not connection state.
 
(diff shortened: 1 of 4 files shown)


pinodeca (approved):
See Waldemort's 2 findings. One looks worth fixing - I'll leave it to you whether to do so in this PR or a followup.

pinodeca (approved):

